PR #14, opened 2026-02-28
fix: resolve SIGABRT/Stack Overflow in IntervalTree via iteration

joergbeyer:
The original implementation used recursive calls for insert, query, and get_mut. 

When data is entered sequentially (e.g., Rows 1, 2, 3...), that creates a unbalanced tree that effectively acts as a linked list. This results in a SIGABRT (Stack Overflow) when the recursion depth exceeds the thread stack limit. Additionally, the default recursive Drop implementation caused crashes during cleanup of large trees.

In addition, this adds a few more test cases.

### DIFF
--- a/crates/formualizer-eval/src/engine/interval_tree.rs
+++ b/crates/formualizer-eval/src/engine/interval_tree.rs
@@ -38,6 +38,20 @@ struct Node<T: Clone + Eq + std::hash::Hash> {
     right: Option<Box<Node<T>>>,
 }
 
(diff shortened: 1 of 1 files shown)


joergbeyer:
I am looking into replacing the IntervallTree with a BTree version. I
have a criterion benchmark running as we speak. I will contribute it, if I
am convinced of the improvement.

I hope you would be open to take a look and consider it.

Joerg

Am Sa., 28. Feb. 2026 um 21:12 Uhr schrieb Frankie Colson <
***@***.***>:

PR #16, opened 2026-02-28
Bench/interval tree performance

joergbeyer:
I added a BTree version of the IntervalTree (side by side in the same source file) and benchmarks to compare. Run "cargo bench" for the benchmark. This will run a bit. No worries, this is only required to compare the 2 different implementations once. 

the BTree version looks convincing to me. 

I also tested that all unit tests pass, if I replace "IntervalTree" with "BTreeIntervalTree" in the 2 files that use the IntervallTree: spill.rs and sheet_index.rs.

PTAL at the benchmark. If you agree, then I can do the following:
 * rename BTreeIntervalTree to IntervalTree (it's an impl, detail what data structure is used), so it will be used in the entire project
 * remove the dup test cases. Currently they test both impls. Only one version is requires
 * simplify the benchmark code, as there is no comparision required going forward. Reduce the number of benchmarks, so it runs a bit faster

### DIFF
--- a/crates/formualizer-eval/benches/interval_tree_bench.rs
+++ b/crates/formualizer-eval/benches/interval_tree_bench.rs
@@ -52,6 +52,17 @@ fn bench_tree_operations(c: &mut Criterion) {
                 BatchSize::SmallInput,
             )
         });
+
+        // --- 5. OVERHEAD RANGE QUERY ---
+        // This specifically tests the "scan risk" noted by the maintainer.
+        // We query for a small window at the very end of a large range.
+        group.bench_with_input(BenchmarkId::new("Query/EndRangeScan", n), n, |b, &n| {
+            b.iter(|| {
+                // Querying for the very last element, but providing a high upper bound
+                // that forces the BTreeMap to return all entries.
+                tree.query(black_box(n - 1), black_box(n))
+            })
+        });
     }
 
     group.finish();
--- a/crates/formualizer-eval/src/engine/interval_tree.rs
+++ b/crates/formualizer-eval/src/engine/interval_tree.rs
@@ -97,14 +97,13 @@ impl<T: Clone + Eq + std::hash::Hash> IntervalTree<T> {
             && let Some(node) = nodes.iter_mut().find(|n| n.high == high)
         {
             let removed = node.values.remove(value);
-            if node.values.is_empty() {
+
+            if removed && node.values.is_empty() {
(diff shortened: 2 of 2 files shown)


joergbeyer:
update: I've completed the full replacement of the BST with the B-Tree implementation in the latest commit.

    API Parity: The IntervalTree and Entry APIs are preserved, making this a drop-in replacement.

    Verified Reliability: All 1,077 tests are green, including stack-safety stress tests.

    Performance: The B-Tree looks confident to me
    
This PR is now ready for a final review as a complete replacement of the original index structure.

PSU3D0:
@joergbeyer Hi — **Manfred here** (Frankie’s AI agent), commenting on Frankie’s behalf.

Thanks for the follow-on work and benchmark additions. I pulled this branch locally and ran tests/clippy/bench. I think this is close, but I’d request a couple changes before merge.

## What I validated locally
- `cargo test -p formualizer-eval` ✅ (1070 passed)
- `cargo test -p formualizer-workbook --test large_sheets` ✅
- `cargo clippy -p formualizer-eval --all-targets -- -D warnings` ✅
- `cargo bench -p formualizer-eval --bench interval_tree_bench ...` ✅

## Blocking correctness issue to fix first
In `interval_tree.rs`, `remove()` currently decrements `self.size` when a **value** is removed, but `size` tracks **interval count**.

That breaks multi-value intervals and can underflow/panic:
- insert `(1,1)->10`
- insert `(1,1)->20`  (same interval)
- remove `10`  -> `len` becomes `0` (already wrong)
- remove `20`  -> `size -= 1` underflows and panics in debug

So this should be fixed before merge.

### Suggested fix
Only decrement `size` when the interval node is actually removed (i.e., value set becomes empty and the interval entry is deleted), not on every successful value removal.

### Suggested tests
Please add a focused test that:
1. inserts two values into the same interval,
2. removes one value and asserts `len() == 1`,
3. removes the second value and asserts `len() == 0` (no panic).

## Perf/readout note (non-blocking but important)
The benchmark is useful, thank you. One thing to tighten in a follow-up: it currently benchmarks only the new structure, so it doesn’t show a direct in-tree baseline comparison by itself.

Also, `query()` now scans `map.range(..=q_high)` which can be effectively full-scan for high/open-ended query bounds; this may still be fine for current scales, but worth explicitly measuring in representative workloads.

---

Once the `remove()/size` bug is addressed and the focused test is in, I’m happy to re-review quickly.

PR #26, opened 2026-03-06
fix(eval): add min_args arity check to prevent panic on too-few arguments

bryanjj:
## Summary

Functions like ROUND (`min_args=2`) could panic with index-out-of-bounds when called with fewer arguments than expected, because per-argument validation only checks declared schema positions.

Adds a `min_args` field to `ValidationOptions` and checks it at the start of `validate_and_prepare`, before any per-argument validation runs. The `dispatch` method now passes `self.min_args()` automatically.

This prevents mutex-poisoning panics in the evaluation engine when formulas have missing arguments.

## Test plan

- [x] 4 new tests: ROUND(1 arg), ROUNDDOWN(1 arg), ABS(0 args), MOD(1 arg) — all return errors instead of panicking
- [x] Updated existing `ValidationOptions` literals in validator.rs
- [x] All existing tests pass

### DIFF
--- a/crates/formualizer-eval/src/args.rs
+++ b/crates/formualizer-eval/src/args.rs
@@ -86,6 +86,11 @@ pub struct PreparedArgs<'a> {
 #[derive(Default)]
 pub struct ValidationOptions {
     pub warn_only: bool,
+    /// Minimum number of arguments the function requires.  When non-zero,
+    /// `validate_and_prepare` rejects calls with fewer arguments before any
+    /// per-argument validation runs, preventing out-of-bounds panics in
+    /// `eval` implementations.
+    pub min_args: usize,
 }
 
 // Legacy adapter removed in clean break.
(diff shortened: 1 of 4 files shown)


NickHilton:
I reviewed the four changed files. The minimum-arity check fixes the reported panic on the normal `Function::dispatch` path, and I found no correctness or security blocker in that path.

**Medium — public API compatibility:** `ValidationOptions` is public, so adding the required `min_args` field breaks downstream code that constructs it with a struct literal such as `ValidationOptions { warn_only: false }`. The updates to the in-repo validator tests illustrate the source break. Could the minimum-arity check run directly in `Function::dispatch` before `validate_and_prepare`? That keeps the fix at the function boundary without changing the public options struct.

Validation: `cargo check -p formualizer-eval --all-targets` and `git diff --check` passed. The eval crate run had 1,097 passing tests and one failure involving three complex-number fixture strings; that failure reproduced identically on the base commit, so it is unrelated to this PR.

PR #36, opened 2026-03-24
feat(eval): implement holiday support for NETWORKDAYS and WORKDAY

TyrGo:
The holidays parameter (third argument) was accepted but silently ignored in both NETWORKDAYS and WORKDAY, producing incorrect results when callers supplied holiday dates.

Changes:
- Add collect_holidays() helper that parses the holidays argument from a cell range, inline array constant, or single scalar value into a HashSet<NaiveDate> for O(1) lookups during day iteration.
- Add literal_to_serial() helper for coercing LiteralValue variants to date serials within flattened holiday arrays.
- Wire both NETWORKDAYS and WORKDAY eval bodies to use collect_holidays(), replacing the empty Vec placeholders.
- Update docstrings and inline examples to reflect correct behavior.
- Add 10 unit tests covering: no holidays, single scalar holiday, array holidays, holiday on weekend (no effect), reversed date range, negative workday offset, holiday on start date, and native Date literal holidays.
- Add 6 JSON formula integration tests for range and scalar holiday inputs via the formula test runner.


TyrGo:
Covered by resent merge. Closing.

Issue #73, opened 2026-04-20
parse: real parser for structured (table) references


PSU3D0:
# parse: real parser for structured (table) references

## Problem

Excel tables use a rich bracketed reference syntax (OOXML / MS-XLSX
§18.17.6.2):

- `Table1[Column]`
- `Table1[[#Headers],[Column]]`
- `Table1[[#All],[ColumnA]:[ColumnB]]`
- `Table1[[@],[Column]]` or `Table1[@Column]`
- `Table1[[#This Row],[Column]]` (legacy Excel 2013/2016 form of `@`)
- `Table1[Col with ']' escape]]` (`]` inside a column name is escaped
  as `'`)
- Case-insensitive specials: `Table1[#headers]` == `Table1[#Headers]`.

The current parser has five compounding defects that cause **silent data
loss** for most non-trivial structured references. For workbooks with
tables (extremely common), this is the single highest-impact parsing bug
in the crate.

### 1. `parse_complex_table_specifier` drops column info

`src/parser.rs:1307-1348` uses naive `content.contains("[#Headers]")`-style
checks, constructs a `Combination` containing only detected `#Special`
items, and **silently discards all columns and column ranges**.

Demonstrations:

- `Table1[[#All],[NPI]:[JMG Group]]` → `Table1[[#All]]` — column range
  lost.
- `Table1[[#Headers],[Column1]:[Column2]]` → `Table1[[#Headers]]`.
- `Table1[[#Data],[Column1]]` → `Table1[[#Data]]` — column dropped.
- `Table1[[@],[Column1]]` → `Column("[@],[Column1]")` raw — both lost.

### 2. Table-column `'`-escape not handled

`parse_brackets` at `src/tokenizer.rs:971-1001` counts `[` and `]` but
does not respect the OOXML `'`-escape rule for column names containing
`]`. `Table1[Col ']']` (column literally named `Col ]`) tokenizes wrong.

### 3. `#This Row` legacy specifier not recognized

`parse_special_item` at `src/parser.rs:1293-1304` only accepts `@` for
current-row. Pre-2019 workbooks serialize `Table1[[#This Row],[Col]]` —
an entire class of workbooks authored by Excel 2013/2016 with tables has
their current-row markers silently dropped.

### 4. `parse_table_reference` accepts trailing garbage

`parse_table_specifier` at `src/parser.rs:1081-1105` finds the first
balanced `]` and parses only that prefix; everything after is silently
[...]

PR #90, opened 2026-04-25
fix(parse): prevent R1C1-shaped input from misclassifying as tables

### LINKED ISSUE #76: parse: prevent R1C1-shaped input from misclassifying as tables
# parse: prevent R1C1-shaped input from misclassifying as tables

## Problem

This issue does **not** add R1C1 dialect support. It only prevents the
current parser from silently misclassifying R1C1-shaped input as a table
reference.

Current behavior: any operand containing `[` is routed through
`parse_table_reference` (`crates/formualizer-parse/src/parser.rs:690-695`).
Input like `=R[1]C[2]` accumulates as a single `Operand:Range` value
`R[1]C[2]`, then `parse_table_reference` splits at the first `[` and
produces:

```
ReferenceType::Table(TableReference {
    name: "R",
    specifier: Some(TableSpecifier::Column("1"))
})
```

Everything after the first balanced `]` is silently discarded (root
cause is the trailing-garbage bug fixed in the structured-references work). The resulting
AST has no connection to R1C1 semantics and evaluating it will produce
wrong answers or `#NAME?`.

`=R1C1` tokenizes as `Operand:Range` and degrades to
`NamedRange("R1C1")` — less catastrophic, but still wrong.

## Out of scope

- Adding a `FormulaDialect::R1C1` variant.
- Parsing R1C1 as actual cell references.
- Converting R1C1 to A1.

## Failing examples

| Input          | Current                                        | Expected (A1 dialect)       |
|----------------|------------------------------------------------|-----------------------------|
| `=R[1]C[2]`    | `Table(TableReference { name: "R", ... })`    | `NamedRange("R[1]C[2]")` or ParserError |
| `=R1C1`        | `NamedRange("R1C1")`                           | unchanged or clean parser error; do **not** misclassify as table |
| `=RC`          | `NamedRange("RC")`                             | Same                        |
| `=R[-1]C`      | likely table misclassification                 | `NamedRange("R[-1]C")` or ParserError |

## Passing (must stay correct)

- All existing table references: `Table1[Column]`, `Table1[[#All],...]`,
  etc. (coordinate with the structured-references work).
- Genuine defined names: `MyName`, `MyTable`, `_xlfn.UNIQUE`.
- Bracketed workbook refs: `[1]Sheet1!A1`.

## Required changes

Two paths:

### Option A: detect R1C1 pattern before table routing

In `parse_excel_reference` (`src/parser.rs:687-699`), add a pre-check:

```rust
if is_r1c1_shape(&ref_part) {
    return Ok(ReferenceType::NamedRange(reference.to_string()));
}
```

Heuristic: `^R\d*(\[\-?\d+\])?C\d*(\[\-?\d+\])?$` AND contains at least
one digit or bracket (so bare `R`/`C`/`RC` are allowed through).

### Option B (recommended): strict trailing-garbage rejection in the structured-references work

The structured-references work adds strict rejection of trailing
content after a valid bracketed specifier. When the suffix after the
first balanced `]` contains more alphabetics or brackets, reject as
invalid table reference. After rejection, fall through to `NamedRange`
classification.

This is the cleaner approach: no R1C1-specific logic needed. This
issue then becomes a regression test ensuring the outcome for
R1C1-shaped input is `NamedRange` rather than a fake table.

### Positive tests

Add to `src/tests/parser.rs` under a new `mod r1c1_disambiguation`:

1. `test_r1c1_bracketed_offsets_not_table` — `=R[1]C[2]` → NOT a
   `Table(TableReference)`. Accepts `NamedRange` or ParserError per
   final design.
2. `test_r1c1_absolute` — `=R1C1` → unchanged named-range behavior or a clean parser error; never a table reference.
3. `test_r1c1_mixed` — `=R1C[2]` → named-range behavior or clean parser error; never a table reference.
4. `test_r1c1_negative_offset` — `=R[-1]C` → named-range behavior or clean parser error; never a table reference.
5. `test_a1_r1_still_cell` — `=R1` remains a normal A1 cell reference (column R, row 1), not a named range.

### Regression tests

1. `=Table1[Col]` unchanged.
2. `=MyDefinedName[3]` (illegal in Excel anyway) → same behavior as
   today or ParserError.
3. `=[1]Sheet1!A1` unchanged (external workbook).

### Tokenizer tests

Confirm tokenizer behavior for `=R[1]C[2]`:
- Currently accumulates as one `Operand:Range` value `R[1]C[2]`.
- May change if the structured-references work's `'`-escape alters bracket handling.

### Cross-parser differential

Both parsers produce identical ASTs for R1C1-shaped input.

## Acceptance criteria

- `=R[1]C[2]` does not produce a `ReferenceType::Table` variant.
- Outcome is either existing named-range behavior or a clear ParserError.
- Existing table reference tests pass.
- No regression in `[1]Sheet1!A1` external-ref path.
- Follow-up R1C1 dialect work is noted as a non-goal.

## Dependencies

- **Benefits from the structured-references work landing first** (strict trailing-garbage
  rejection). This issue then becomes mostly a regression test.

## Notes

- Excel's OOXML files store formulas in A1 form exclusively. R1C1 is a
  UI display mode. R1C1-shaped strings should not appear in stored
  formulas.
- Real R1C1 support would require a new `FormulaDialect::R1C1` and a
  distinct parsing path; out of scope.

PSU3D0:
Closes #76

## Problem

Operands like `R1C[2]`, `RC[1]`, `R5C[1]` contain `[` and were routed
through `parse_table_reference`, producing fake
`Table { name: \"R1C\", specifier: Column(\"2\") }` ASTs unrelated to
R1C1 semantics. Forms with bracketed row offsets (`R[1]C[2]`,
`R[-1]C`) hit the structured-references trailing-garbage check (#73)
and surfaced as parser errors, but only by accident.

This PR is **not** R1C1 dialect support. OOXML stores formulas in A1
only; we just refuse to fabricate a fake `Table` AST and route
R1C1-shaped operands to the same `NamedRange` outcome that bracket-
free `R1C1`/`RC` already produce.

## Fix

`parse_excel_reference` adds an `is_r1c1_shape` gate ahead of the
table path. The shape detector matches
`^R\\d*(\\[-?\\d+\\])?C\\d*(\\[-?\\d+\\])?\$` and additionally requires
at least one digit or bracket so that bare `R`, `C`, `RC` fall
through untouched. Plain A1 cells like `R1`, `C5`, `RC1` never reach
this path because they don't contain `[`.

Stacked on `parse/issue-73-structured-refs` (real recursive-descent
specifier parser); base is set accordingly.

## Tests

New `r1c1_disambiguation` module in `crates/formualizer-parse/src/tests/parser.rs`:

- `test_r1c1_bracketed_offsets_not_table` — `=R[1]C[2]` is not a Table
- `test_r1c1_absolute` — `=R1C1` stays NamedRange
- `test_r1c1_mixed` — `=R1C[2]` is not a Table (regression)
- `test_r1c1_negative_offset` — `=R[-1]C` is not a Table
- `test_r1c1_rc_with_bracket_only_col` — `=RC[1]` is not a Table (regression)
- `test_r1c1_row_digit_col_bracket` — `=R5C[1]` is not a Table (regression)
- `test_a1_r1_still_cell` — `=R1` remains an A1 Cell
- `test_a1_c5_still_cell` — `=C5` remains an A1 Cell
- `test_table_reference_unchanged` — `Table1[Col]` still Table
- `test_external_workbook_ref_unchanged` — `[1]Sheet1!A1` still External
- `test_cross_parser_agreement_r1c1` — both classic and span-based
  parsers agree (no Table) for the full R1C1-shape battery

## Validation

- \`cargo fmt --all\`
- \`cargo test -p formualizer-parse r1c1_disambiguation\` — 11/11 pass
- \`cargo test -p formualizer-parse\` — 195 lib + 3 integration pass
- \`cargo clippy -p formualizer-parse --all-targets -- -D warnings\` — clean

### DIFF
--- a/crates/formualizer-parse/src/parser.rs
+++ b/crates/formualizer-parse/src/parser.rs
@@ -708,6 +708,16 @@ impl ReferenceType {
         // Table references live in the ref_part (e.g., "Table1[Column]").
         // Sheet names can contain '[' for external workbook refs (e.g., "[1]Sheet1!A1").
         if ref_part.contains('[') {
+            // Issue #76: R1C1-shaped operands like `R[1]C[2]`, `R1C[2]`, `RC[1]`
+            // contain `[` but are not table references. Without this gate they
+            // either misclassify as `Table { name: "R1C", specifier: Column("2") }`
+            // or get rejected by the structured-references trailing-garbage check.
+            // We don't add an R1C1 dialect; we just refuse to fabricate a table
+            // and fall back to the same `NamedRange` outcome that bracket-free
+            // R1C1 strings (e.g. `R1C1`, `RC`) already produce.
+            if Self::is_r1c1_shape(&ref_part) {
+                return Ok(ReferenceType::NamedRange(reference.to_string()));
+            }
             return Self::parse_table_reference(&ref_part);
         }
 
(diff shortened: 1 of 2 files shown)


PSU3D0:
Integrated into main via the parser merge train (main now includes this PR's head commits; final integration head: 7a52e8f69468cd6c7f75112be5bf4aa2b491a4a3). Closing this stacked PR because GitHub cannot retarget it after the head became empty against main.

PR #106, opened 2026-06-02
fix(eval): drain all staged sheets in evaluate_cell / evaluate_cells

bryanjj:
## Problem
`evaluate_cell` / `evaluate_cells` only drained the *target* sheet's staged cells before evaluating. Cross-sheet references to formula cells living on a different staged sheet therefore resolved to empty/`None` instead of their computed value.

## Fix
Drain all staged sheets (not just the target) before evaluation so cross-sheet references resolve correctly.

## Test
Adds `cross_sheet_named_range_first_cell` regression test. Verified the test **fails on current `main`** without the fix (`evaluate_cell must drain all staged sheets ...; got None`) and passes with it.

### DIFF
--- a/crates/formualizer-eval/src/engine/eval.rs
+++ b/crates/formualizer-eval/src/engine/eval.rs
@@ -8689,8 +8689,15 @@ where
                 .with_message("Row and column must be >= 1".to_string()));
         }
 
+        // ``defer_graph_building`` mode stages formulas during bulk load
+        // and lazily promotes them into the dependency graph at evaluate
+        // time. Per-cell evaluation must drain *all* staged sheets, not
+        // just the requested target — a cell's formula can reference
+        // any sheet in the workbook, and a cross-sheet ref to a still-
+        // staged source would silently evaluate to ``None`` if that
+        // source sheet hadn't been promoted yet.
         if self.config.defer_graph_building {
-            self.build_graph_for_sheets(std::iter::once(sheet))?;
+            self.build_graph_all()?;
         }
 
         let result = self.evaluate_cells(&[(sheet, row, col)])?;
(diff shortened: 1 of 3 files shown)


PSU3D0 (approved):
Reviewed and reproduced the bug. The failing case on current `main` is specifically the deferred-graph path: `evaluate_cell` promotes only the target sheet, then a cross-sheet reference to a formula cell on a still-staged sheet resolves as `None`.

I initially worried that `build_graph_all()` was too broad because it collapses the per-sheet laziness of `defer_graph_building` (e.g. first `evaluate_cell` on one sheet in a many-sheet workbook now pays the full graph-build cost). But after looking at the dependency model, I think this is the correct conservative fix for 0.6:

- staged formulas are raw text; their dependency edges don't exist until promotion;
- cross-sheet refs are only the simple case;
- dynamic/opaque dependencies (`INDIRECT`, computed refs, names resolving across sheets, 3D refs, future table/external ref handling, etc.) make pre-promotion sheet-closure discovery incomplete unless we duplicate a large part of the real dependency resolver;
- partial per-sheet graph materialization is therefore semantically unsound for demand evaluation unless it has a very conservative fallback.

So this restores the safest invariant: before demand-driven evaluation, a deferred workbook has a globally complete dependency graph.

Performance note: this doesn't add work compared with eager graph-building; it moves the full build to the first evaluation. It does remove the current "peek one sheet without materializing the rest" optimization. A future optimization could add a static sheet-closure fast path, but it should always fall back to `build_graph_all()` for opaque/dynamic references.

Validation on the PR branch:

- `cargo test -p formualizer-eval` → 1749 passed / 0 failed
- `cargo clippy -p formualizer-eval --all-targets -- -D warnings` → clean

Approving as the correctness-first fix for 0.6.

PR #107, opened 2026-06-02
fix(eval): route date-function coercion through to_number_lenient; EDATE/EOMONTH negative-month boundary

bryanjj:
Two related date-function fixes.

## 1. Route date-function coercion through `to_number_lenient`
Date functions (date-part extraction, `EDATE`/`EOMONTH`, `WEEKDAY`/`WORKDAY`) did not coerce `Date`/`DateTime` cell values to serial numbers, returning `#VALUE!` when given a real date cell instead of a numeric serial. Routing coercion through `to_number_lenient` makes them accept date-typed cells.

## 2. EDATE/EOMONTH year-boundary off-by-one with negative months
`EDATE`/`EOMONTH` computed the wrong month when a negative month offset crossed a year boundary.

## Tests
Adds `edate_eomonth_engine` and `date_function_duration_cells` engine tests plus `date_time.json` cases. Verified the new engine tests **fail on current `main`** (e.g. `EDATE` with a date cell returns `#VALUE!`) and pass with the fix.

### DIFF
--- a/crates/formualizer-eval/src/builtins/datetime/date_parts.rs
+++ b/crates/formualizer-eval/src/builtins/datetime/date_parts.rs
@@ -1,6 +1,6 @@
 //! Date and time component extraction functions
 
-use super::serial::{date_to_serial, datetime_to_serial, serial_to_date, serial_to_datetime};
+use super::serial::{serial_to_date, serial_to_datetime};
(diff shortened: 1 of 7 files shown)


PSU3D0 (approved):
Good addition

PR #108, opened 2026-06-02
fix(eval): named-range pass-through vertices in demand subgraph + preserve edges through CSR rebuild

bryanjj:
Two related fixes for named-range dependency tracking during incremental evaluation.

## 1. Walk through Named/Range pass-through vertices in the demand subgraph
`evaluate_cell`'s demand subgraph did not walk *through* `Named`/`Range` pass-through vertices, so dirty upstream cells sitting behind a named range were not recomputed. After overriding an input, the target kept a stale value.

## 2. Preserve named-range edges through CSR rebuild + schedule named pass-through vertices
The CSR rebuild dropped named-range edges, and named pass-through vertices were not scheduled, so xlsx-ingested defined-name ranges produced wrong results after `set_value`.

## Tests
Adds `demand_subgraph_named_range` and `dn_range_xlsx_subgraph` regression tests. Verified both **fail on current `main`** (stale values after override) and pass with the fix.

### DIFF
--- a/crates/formualizer-eval/src/engine/delta_edges.rs
+++ b/crates/formualizer-eval/src/engine/delta_edges.rs
@@ -417,6 +417,21 @@ impl DeltaEdgeSlab {
         self.coord_changed = false;
     }
 
+    /// Iterate over all (from, &additions) pairs in the delta. Used by
+    /// build_from_adjacency to carry forward delta-only edges that the
+    /// adjacency input does not cover.
+    pub fn additions_iter(&self) -> impl Iterator<Item = (&VertexId, &FxHashSet<VertexId>)> {
+        self.additions.iter()
+    }
+
+    /// Return removals scheduled for `from` (empty slice if none).
+    pub fn removals_for(&self, from: VertexId) -> impl Iterator<Item = VertexId> + '_ {
+        self.removals
+            .get(&from)
+            .into_iter()
+            .flat_map(|set| set.iter().copied())
+    }
+
     /// Apply delta to CSR, creating a new CSR structure
     pub fn apply_to_csr(
         &self,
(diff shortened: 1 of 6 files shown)


PSU3D0:
Reviewed the correctness and ran a before/after load benchmark. Summary: the fix is correct and there's **no load regression**, but I'd like one cleanliness change before merge.

### Correctness ✅
All three changes are justified and the two new regression tests fail on `main` and pass here. Full `formualizer-eval` suite (1748 passed / 0 failed) and clippy are clean on the branch.

### Performance ✅ (no regression)
`build_from_adjacency` has exactly one non-test caller — the one-shot bulk-ingest finalize in `ingest_builder.rs`. Incremental edits go through `rebuild()` → `apply_to_csr` (the delta path), which this PR does **not** touch. So the carry-forward loop only runs once at load finalize, never per-edit.

Load+eval, 50k rows × 16 cols, identical fixtures, median of 3 runs:

| Scenario | main load | this PR load | Δ |
|---|---|---|---|
| linear_rollup | 2597.7 ms | 2584.6 ms | −0.5% |
| sumifs_report | 1846.3 ms | 1832.9 ms | −0.7% |
| whole_column_summary | 443.4 ms | 446.9 ms | +0.8% |

All within run-to-run noise.

### Requested change 🧹
`build_from_adjacency(mut adjacency, ...)` now silently changed its contract. It used to mean "replace edges from exactly this adjacency." It now also reads `self.base`/`self.delta`/`self.vertex_ids` and merges in carry-forward edges for uncovered vertices. The `mut` on the param is the visible symptom of that hidden semantic shift (it's memory-safe since the `Vec` is owned by value, but it's surprising).

Could you keep `build_from_adjacency` a **pure builder** — either:
- **(preferred)** compute the merged adjacency in the caller (or a clearly named helper) and pass the final `Vec` in, so the signature stays `adjacency: Vec<(u32, Vec<u32>)>` with no `mut` and no internal `self.*` reads; or
- rename it to make the behavior explicit (e.g. `build_from_adjacency_preserving_uncovered`) and document the contract that it merges in existing edges for vertices absent from the input.

The new `additions_iter` / `removals_for` accessors are fine to keep either way.

bryanjj:
thanks for the review. made the recommended changes. let me know if anything else

PSU3D0 (approved):

PR #152, opened 2026-06-19
fix: resolve sheet names case-insensitively (Excel-compatible)

verticka:
## Problem

Sheet-name resolution is case-sensitive, which differs from Excel. For example,
`INDIRECT("Config!A1")` returns `#REF!` when the sheet is actually named `CONFIG`.

This breaks workbooks that build references at runtime, e.g.
`INDIRECT(ADDRESS(row, col, 1, 1, "Config"))`, whenever the casing of the
sheet-name string differs from the stored sheet name. Excel resolves sheet
names case-insensitively (and they are unique case-insensitively).

## Fix

`SheetRegistry` now keys `id_by_name` by the lowercased sheet name, while
`name_by_id` keeps the original casing for display. `id_for`, `get_id`,
`remove` and `rename` all match regardless of casing.

## Test

Added a unit test asserting that `CONFIG` / `Config` / `config` resolve to the
same sheet id, that `id_for` does not create a duplicate sheet, and that the
original casing is preserved.

### DIFF
--- a/crates/formualizer-eval/src/engine/sheet_registry.rs
+++ b/crates/formualizer-eval/src/engine/sheet_registry.rs
@@ -14,13 +14,16 @@ impl SheetRegistry {
     }
 
     pub fn id_for(&mut self, name: &str) -> SheetId {
-        if let Some(&id) = self.id_by_name.get(name) {
+        // Sheet names are CASE-INSENSITIVE (Excel behavior): index by the lowercased name, but
+        // keep the original casing in name_by_id for display.
+        let key = name.to_lowercase();
+        if let Some(&id) = self.id_by_name.get(&key) {
             return id;
         }
 
         let id = self.name_by_id.len() as SheetId;
         self.name_by_id.push(name.to_string());
-        self.id_by_name.insert(name.to_string(), id);
+        self.id_by_name.insert(key, id);
         id
     }
 
@@ -33,7 +36,8 @@ impl SheetRegistry {
     }
 
     pub fn get_id(&self, name: &str) -> Option<SheetId> {
-        self.id_by_name.get(name).copied()
+        // Case-insensitive (Excel): e.g. INDIRECT("Config!B8") must find the sheet named "CONFIG".
+        self.id_by_name.get(&name.to_lowercase()).copied()
     }
 
     /// Count active sheets without cloning sheet names.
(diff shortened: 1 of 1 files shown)


PSU3D0:
Thanks for this — the bug is real and the fix is correct. I verified it across INDIRECT, cross-sheet refs and 3D spans, and it also closes the case where two sheets differing only by case could both exist, which Excel doesn't allow.

I'm going to hold it though, because I want to land sheet case-folding slightly differently. Names and tables already fold through config-gated helpers, and this adds a fourth, hardcoded fold in `SheetRegistry` — I'd rather have one shared case-fold that sheets, names and tables all go through, which also lets us pin down the Unicode-vs-ASCII behaviour (Turkish-i and friends) in one place instead of three. Your diagnosis is what's driving that, and I'll credit it. I'll link the design back here once it's written.

PSU3D0 (changes requested):
The lookup-path fix is right — INDIRECT and cross-sheet references should resolve sheet names case-insensitively, and keying `id_by_name` by the lowercased name while `name_by_id` keeps display casing is the correct shape. The registry unit test is good too. But it can't merge as-is: it turns case-only renames into silent data loss.

**Blocker: case-only rename desyncs the registry from the Arrow store.** In `graph::rename_sheet` (`crates/formualizer-eval/src/engine/graph/sheets.rs`, ~line 253) the duplicate check calls `get_id(new_name)`. Rename `"config"` → `"CONFIG"`: with your case-insensitive `get_id` that now returns the *same* sheet's id, so `existing_id == sheet_id` and the function early-returns `Ok(())` without ever calling `sheet_reg.rename`. But the caller (`eval.rs::rename_sheet`, ~line 2108) has *already* renamed the sheet in the Arrow store to `"CONFIG"` and sees `Ok`, so nothing rolls back. The registry's canonical name stays `"config"` while the Arrow store's sheet is now `"CONFIG"` — and `SheetStore::sheet()` is case-sensitive by canonical name, so every subsequent read of that sheet resolves id → `"config"` → miss → empty range. All data on the sheet becomes unreadable and the display name never updates. Pre-PR this worked, because the case-sensitive `get_id("CONFIG")` returned `None` and the rename proceeded.

Fix: in `graph::rename_sheet`, when `existing_id == sheet_id` but `sheet_reg.name(id) != new_name`, don't early-return — fall through and call `sheet_reg.rename` (your updated `SheetRegistry::rename` already handles the casing update correctly; it just never gets called). Please add a test that renames a sheet to a different casing of itself and then reads a cell back through the engine.

Non-blocking follow-ups I'll file separately (flagging so you don't feel obliged to fix here): (1) loaders that can produce case-colliding sheet names (JSON backend; xlsx can't) now silently orphan the second sheet — `add_sheet` reuses the first id but `ensure_arrow_sheet` is case-sensitive and pushes a second Arrow sheet; (2) `get_id`/`id_for` now allocate via `to_lowercase()` on the per-reference eval hot path — worth a borrowed-key or cached-id scheme eventually; (3) Rust `to_lowercase()` is Unicode-default folding, not Excel's — fine in practice and consistent with how named ranges are handled, just noting.

PR #153, opened 2026-06-20
fix(FIND/SEARCH): index by character, not byte (UTF-8 safe)

verticka:
## Problem

`FIND` and `SEARCH` index `within_text` by **byte** offset and return **byte** positions. On multi-byte UTF-8 input this is both incorrect and unsafe:

- **Panic.** `SEARCH`'s wildcard scan slices `&hay[offset..]` at every byte offset. When an offset falls inside a multi-byte character it panics:
  ```
  byte index N is not a char boundary; it is inside 'é' (bytes ..)
  ```
  A single accented character in `within_text` can crash a whole evaluation.
- **Wrong positions.** `FIND`/`SEARCH` return Excel positions, which are **1-based character** positions. The current code returns byte offsets, so any match after a non-ASCII character is off by the extra bytes.
- **`?` wildcard.** In `SEARCH`, `?` matched a single **byte** instead of a single **character**.

Minimal repro:
```
=SEARCH("?z","éz")   ' panics (offset lands inside 'é')
=FIND("z","éz")      ' returns 3 (byte), Excel says 2 (char)
```

## Fix

Rework the matching in **character space** (`char_find`, `char_wildcard_search`, `wildcard_match_chars`):

- `FIND`/`SEARCH` return 1-based **character** positions.
- `?` matches exactly one **character**.
- No slicing at non-char boundaries → no panic on accented / non-ASCII text.

A1-style/ASCII behaviour is unchanged; only the indexing model is corrected.

## Test

Added `find_search_utf8_char_positions` covering character positions and the previously-panicking wildcard scan over accented text.

Found on real-world spreadsheets whose text labels contain French accents.

### DIFF
--- a/crates/formualizer-eval/src/builtins/text/find_search_exact.rs
+++ b/crates/formualizer-eval/src/builtins/text/find_search_exact.rs
@@ -445,16 +445,17 @@ mod tests {
         let ctx = wb.interpreter();
         let f = ctx.context.get_function("", "FIND").unwrap();
         let s = ctx.context.get_function("", "SEARCH").unwrap();
-        let call = |func: &std::sync::Arc<dyn crate::function::Function>, needle: &str, hay: &str| {
-            let n = lit(LiteralValue::Text(needle.into()));
-            let h = lit(LiteralValue::Text(hay.into()));
-            func.dispatch(
-                &[ArgumentHandle::new(&n, &ctx), ArgumentHandle::new(&h, &ctx)],
-                &ctx.function_context(None),
-            )
-            .unwrap()
-            .into_literal()
-        };
+        let call =
+            |func: &std::sync::Arc<dyn crate::function::Function>, needle: &str, hay: &str| {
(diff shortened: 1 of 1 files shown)


PSU3D0 (approved):
Nice catch on the panic — confirmed `SEARCH("?z","éz")` blows up on a char boundary on main, and the positions were byte-based rather than char-based too. The rewrite into char space looks right: FIND stays literal and case-sensitive, SEARCH keeps wildcards and case-insensitivity, and `start_num` counts characters now. Local tests pass.

A few adjacent things still worth doing at some point, none of them on you since they predate this: SEARCH doesn't handle the `~*`/`~?` escapes, empty `find_text` returns 1 instead of `start_num`, and positions count Rust chars rather than UTF-16 units so emoji/astral text stays off by one. I'll note those separately. Thanks for the fix.

PSU3D0:
@verticka Just a small formatting issue

PR #154, opened 2026-06-20
fix(INDIRECT): resolve defined names/tables when a1_style is FALSE

verticka:
## Problem

`INDIRECT(text, a1_style)` returns `#N/IMPL!` for **any** call with `a1_style = FALSE`, before attempting to resolve `text`:

```rust
if !a1_style {
    return Some(Err(ExcelError::new(ExcelErrorKind::NImpl).with_message(
        "INDIRECT with R1C1 style (second argument FALSE) is not yet supported",
    )));
}
```

That is wrong for **defined names** and **tables**. The A1/R1C1 distinction only applies to literal cell/range *addresses* (`"A1"` vs `"R1C1"`). A defined name like `"MyRange"` or a structured table reference is neither A1 nor R1C1 syntax, so Excel resolves `INDIRECT("MyRange", FALSE)` exactly like `INDIRECT("MyRange")`.

Refusing it with `#N/IMPL!` breaks real workbooks that dereference a name while defensively passing the R1C1 flag, e.g.:
```
INDIRECT("SomeName" & Sheet!E39, 0)
```

## Fix

In the `a1_style = FALSE` branch, first try to interpret `text` as a named range or table and return it as-is; only genuine R1C1 cell/range text remains unsupported (still `#N/IMPL!`). A1-style behaviour is unchanged.

## Test

Replaced the obsolete `indirect_a1_false_returns_not_implemented_error` (which asserted the old blanket behaviour) with `indirect_a1_false_resolves_named_range`, verifying `INDIRECT(name, FALSE)` resolves the defined name like `INDIRECT(name)`.

### DIFF
--- a/crates/formualizer-eval/src/builtins/reference_fns.rs
+++ b/crates/formualizer-eval/src/builtins/reference_fns.rs
@@ -669,9 +669,17 @@ impl Function for IndirectFn {
         };
 
         if !a1_style {
-            return Some(Err(ExcelError::new(ExcelErrorKind::NImpl).with_message(
-                "INDIRECT with R1C1 style (second argument FALSE) is not yet supported",
-            )));
+            // The A1/R1C1 flag does not apply to defined names or tables (they are
+            // neither A1 nor R1C1 syntax). Excel resolves `INDIRECT(name, FALSE)`
+            // exactly like `INDIRECT(name)`, so handle those before refusing R1C1.
+            // Real R1C1 cell/range text remains unsupported.
+            return match formualizer_parse::parser::ReferenceType::from_string(&ref_text) {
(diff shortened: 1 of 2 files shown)


PSU3D0 (approved):
Confirmed main returns `#N/IMPL!` for defined names under FALSE, which is wrong — the A1/R1C1 flag doesn't apply to names or tables. The fix runs them through the same resolution as the default path, undefined names still come back as `#REF!`, and tables resolve either way. Looks good.

The doc comment above the function still says `a1_style=FALSE` is unimplemented and returns `#N/IMPL!` — that's only true for literal R1C1 addresses now, so I'll fix that wording when I bring this in. (We still don't actually interpret R1C1 addressing — those strings get treated as names, issue #76 — but that's separate from what you're fixing.)

PR #155, opened 2026-06-20
fix(TEXT): return text unchanged for non-numeric input (Excel parity)

verticka:
## Summary

Excel's `TEXT()` returns its first argument **unchanged** when that argument is text that cannot be parsed as a number. A numeric format string does not coerce arbitrary text, and `TEXT` never raises `#VALUE!` for a text argument.

formualizer instead returned `#VALUE!` for non-numeric text. Worse, when that result was concatenated, the error surfaced *inside* the generated string, silently corrupting downstream lookups.

## Reproduction (open a blank sheet, enter the formula, recalc)

| Formula | Excel | formualizer (before) |
|---|---|---|
| `=TEXT("abc","00")` | `abc` | `#VALUE!` |
| `=TEXT("3-1","00")` | `3-1` | `#VALUE!` |
| `=TEXT(MID("x",5,3),"00")` | *(empty)* | `#VALUE!` |
| `="V"&TEXT("3-1","00")` | `V3-1` | `V#VALUE!` |
| `=TEXT("5","00")` | `05` | `05` (already ok) |
| `=TEXT(12.34,"0.00")` | `12.34` | `12.34` (already ok) |

## Fix

In `TextFn::eval`, when the text argument is not a parseable number, return it unchanged instead of `#VALUE!`. Numeric and numeric-text arguments keep formatting as before.

A regression test (`text_non_numeric_text_passes_through`, covering `abc` / `3-1` / `10-`) is included.

### DIFF
--- a/crates/formualizer-eval/src/builtins/text/value_text.rs
+++ b/crates/formualizer-eval/src/builtins/text/value_text.rs
@@ -298,14 +298,14 @@ impl Function for TextFn {
         let num = match val {
             LiteralValue::Number(f) => f,
             LiteralValue::Int(i) => i as f64,
-            LiteralValue::Text(t) => {
-                let Some(n) = ctx.locale().parse_number_invariant(&t) else {
-                    return Ok(crate::traits::CalcValue::Scalar(LiteralValue::Error(
-                        ExcelError::new_value(),
-                    )));
-                };
-                n
-            }
+            LiteralValue::Text(t) => match ctx.locale().parse_number_invariant(&t) {
+                Some(n) => n,
+                // Excel: TEXT() on text that is not a parseable number returns the
+                // text unchanged (e.g. =TEXT("abc","00") -> "abc"), it does not error.
+                None => {
+                    return Ok(crate::traits::CalcValue::Scalar(LiteralValue::Text(t)));
+                }
+            },
             LiteralValue::Boolean(b) => {
                 if b {
                     1.0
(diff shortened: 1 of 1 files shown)


PSU3D0 (changes requested):
The core idea is right — `TEXT` shouldn't throw `#VALUE!` on plain text like `"abc"`. But it can't go in as written.

It breaks an existing test first off: `test_text_formatting` pins `TEXT("1.234,56","0.00")` to `#VALUE!`, and the comment there is deliberate — locale-ambiguous text should error, not silently parse. The passthrough returns it unchanged, so that one fails.

The boundary is the bigger thing though. You pass through anything that isn't a plain number, but Excel coerces text that looks like a date, currency or fraction — `"3-1"`, `"$5"`, `"1/2"` would get formatted in Excel rather than passed through. I'm fairly sure `"3-1"` lands as a date, which would make the `"3-1" → "3-1"` case in your test wrong; worth checking in real Excel before pinning it. Doing this properly probably means routing `TEXT` and `VALUE` through one shared coercion that understands dates/currency/fractions, which is a larger chunk of future work. For now I'd keep it narrow: only pass through clearly non-numeric text (keep `"abc"`, drop `"3-1"`/`"10-"`) and leave the `"1.234,56"` case alone.

Separate heads-up: the format engine doesn't pad leading zeros, so `TEXT(5,"00")` already returns `"5"` not `"05"` — the description's examples won't render that way regardless.

PSU3D0 (approved):
Yeah, this is the right shape — gating on whether the text has a digit keeps dates/currency/fractions as `#VALUE!` and leaves the `"1.234,56"` case alone, while `"abc"`/`"N/A"` pass through like Excel. Ran it locally, `test_text_formatting` is green again and the two new tests cover both sides. Approving, thanks for the quick turnaround.

The shared TEXT/VALUE coercion is still the real long-term fix for the digit-bearing cases, but that's its own piece of work — no reason to hold this up for it.

PR #156, opened 2026-06-20
fix(INDEX): support row_num/column_num = 0 (entire row/column)

verticka:
## Summary

In Excel, passing `0` as `row_num` or `column_num` to `INDEX` selects a whole line of the array:

- `INDEX(range, r, 0)` → the **entire row** `r`
- `INDEX(range, 0, c)` → the **entire column** `c`
- `INDEX(range, 0, 0)` → the **entire range**

This is a common idiom inside aggregates (`=SUM(INDEX(a, r, 0))`) and when feeding a whole row/column to another function.

formualizer rejected a `0` index with `#REF!` because of the guard `if row <= 0 || col <= 0`.

## Reproduction

| Formula (B1:B5 = 10,20,30,40,50; A1:C3 rows) | Excel | formualizer (before) |
|---|---|---|
| `=INDEX(B1:B5,2,0)` | `20` | `#REF!` |
| `=INDEX(A1:C1,0,2)` | `20` | `#REF!` |
| `=SUM(INDEX(A1:C3,2,0))` | sum of row 2 | `#REF!` |
| `=SUM(INDEX(A1:C3,0,2))` | sum of column 2 | `#REF!` |
| `=INDEX(A1:C3,-1,2)` | `#REF!` | `#REF!` (unchanged) |

## Fix

In `IndexFn::eval_reference`, treat a `0` index as "entire row/column":

- `column_num == 0` → return the row range `(r, sc)..(r, ec)` (or the single cell if the source is one column),
- `row_num == 0` → return the column range `(sr, c)..(er, c)` (or the single cell if the source is one row),
- both `0` → the entire range,
- negative indices remain `#REF!`.

Adds regression tests for the degenerate single-cell cases, full row/column materialization via `SUM`, and the negative-index `#REF!` case.

### DIFF
--- a/crates/formualizer-eval/src/builtins/reference_fns.rs
+++ b/crates/formualizer-eval/src/builtins/reference_fns.rs
@@ -229,9 +229,52 @@ impl Function for IndexFn {
             }
         };
 
-        // 1-based indexing per Excel
-        if row <= 0 || col <= 0 {
+        // 1-based indexing per Excel. A 0 means "the entire row" (column_num == 0)
+        // or "the entire column" (row_num == 0); 0 for both yields the whole range.
+        // Negative indices are #REF!.
+        if row < 0 || col < 0 {
             return Some(Err(ExcelError::new(ExcelErrorKind::Ref)));
+        }
+        let range_ref = |sheet, sr, sc, er, ec| ReferenceType::Range {
+            sheet,
+            start_row: Some(sr),
+            start_col: Some(sc),
+            end_row: Some(er),
+            end_col: Some(ec),
+            start_row_abs: false,
+            start_col_abs: false,
+            end_row_abs: false,
+            end_col_abs: false,
+        };
+        if col == 0 {
+            if row == 0 {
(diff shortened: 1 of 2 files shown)


PSU3D0 (changes requested):
Reference path checks out — 0 for row/col gives the whole row or column, negatives and out-of-bounds still return `#REF!`, and summing over the result works. That part's solid.

It's incomplete though: the array path (`IndexFn::eval`, the `index <= 0` and `row == 0 || col == 0` guards around lines 345 and 378) still rejects 0, so `INDEX({1,2,3},0)` and `INDEX({1,2;3,4},0,2)` over array constants come back `#REF!` where Excel gives the whole row/column. Can you extend the same handling there and add a test for the array-constant case so the two paths don't drift apart? Then it's good to go.

PSU3D0 (approved):
Both paths handle 0 now, and the array-constant cases are all covered — `INDEX({1,2,3},0)`, the `0,0` whole-array, the single-row/col degenerations, and negatives still `#REF!`. Ran the index suite locally, all green. Approving.

PR #157, opened 2026-06-24
Batch workbook formula cache recalculation

ohnoah:
## Summary
- add a Rust-side Python binding entrypoint, `Workbook.recalculate_formula_cells`, for batch formula target evaluation
- update `recalculate_file` to collect workbook formula targets and evaluate normal workbooks in one Rust batch before writing XLSX caches
- preserve the existing per-cell `_xlfn` fallback path for workbooks containing `_xlfn` formulas so dependent formula behavior does not regress
- add direct batch API and `_xlfn` dependent regression coverage

## Verification
- `uv run pytest tests/test_recalculate_file.py tests/test_recalculate_file_extensive.py` -> 13 passed, 1 upstream fixture warning
- synthetic binding benchmark: 10,000 formula targets, repeated `evaluate_cell` 0.0831s vs batch `recalculate_formula_cells` 0.0433s (~1.92x)
- synthetic end-to-end `recalculate_file`: 20,000 formula workbook, 0 errors, 1.1642s, cached values verified

## Notes
- `cargo test -p formualizer-python` is not valid in this local setup: the PyO3 dylib target fails to link Python symbols directly. The `uv run pytest` path builds/imports the extension successfully.
- `Cargo.lock` had stale 0.3.6 versions for manifests that are already 0.3.7 on main; Cargo corrected those entries.


ohnoah:
Closing: this branch targets fork-only Python recalculate_file code that exists on ohnoah/formualizer but not PSU3D0/main. Re-opening against ohnoah/formualizer, which is the package source Newton consumes.

PR #159, opened 2026-07-01
fix(criteria): count blank cells for <> criteria in SUMIFS/COUNTIFS (bounded to used region)

### LINKED ISSUE #158: SUMIFS/SUMIF/COUNTIFS "<>value" criteria exclude blank cells (Excel includes them)
**Version:** formualizer 0.7.0 (PyPI, Python bindings)

### Summary

A not-equal criterion (`"<>X"`) in `SUMIFS` / `SUMIF` / `COUNTIFS` excludes empty/blank cells from the match. In Excel and LibreOffice a blank cell **satisfies** a `<>X` criterion (blank ≠ `"X"` is TRUE), so blanks are counted. `formualizer` omits them, which silently under-counts any aggregation that excludes a category via `<>` when the criteria column contains blank cells.

### Minimal reproduction

```python
import formualizer as fz

wb = fz.Workbook()
s = wb.sheet("S")
s.set_value(1, 1, "Debt");   s.set_value(1, 2, 10)
# A2 deliberately left blank
s.set_value(2, 2, 20)
s.set_value(3, 1, "Equity"); s.set_value(3, 2, 30)

s.set_formula(1, 4, '=SUMIFS(B1:B3, A1:A3, "<>Debt")')
s.set_formula(2, 4, '=COUNTIFS(A1:A3, "<>Debt")')
s.set_formula(3, 4, '=SUMIF(A1:A3, "<>Debt", B1:B3)')

print(wb.evaluate_cell("S", 1, 4))  # SUMIFS
print(wb.evaluate_cell("S", 2, 4))  # COUNTIFS
print(wb.evaluate_cell("S", 3, 4))  # SUMIF
```

### Expected (Excel / LibreOffice)

```
SUMIFS   = 50    # blank A2 (20) + Equity (30)
COUNTIFS = 2     # blank A2 + Equity
SUMIF    = 50
```

### Actual (formualizer 0.7.0)

```
SUMIFS   = 30    # only the explicit "Equity" row; the blank A2 is dropped
COUNTIFS = 1
SUMIF    = 30
```

### Notes

- Reproduces via both `evaluate_cell` and `recalculate_file`.
- `SUM` over the same range is correct — only the `<>` criterion matching is affected.
- Equality criteria (e.g. `"Equity"`) and other comparisons are unaffected; the issue is specifically that a blank cell should satisfy a `<>` predicate.
- Real-world impact: sheets exported from other systems often leave a category column blank for most rows, so `SUMIFS(value, category, "<>Debt", category, "<>Commodity")` silently drops every blank-category row.

### Reference

Microsoft `SUMIFS`/`COUNTIFS` semantics — a `"<>text"` criterion matches empty cells.

PSU3D0:
Fixes #158.

## Problem

A `"<>X"` criterion in SUMIFS/SUMIF/COUNTIFS/AVERAGEIFS silently excluded blank cells. In Excel and LibreOffice a blank equals the empty string, so a blank satisfies `<>X` (blank ≠ "X" is TRUE) and must be counted. Any aggregation that excludes a category via `<>` over a column with blank rows was under-counting.

## Root cause

`compute_criteria_mask` (the Arrow-vectorized cached criteria-mask builder in `crates/formualizer-eval/src/engine/eval.rs`) evaluates `<>X` text criteria with Arrow's `nilike`. Arrow follows SQL null semantics: `NULL NOT ILIKE 'x'` yields NULL, not TRUE. Blank/Empty cells are NULL in the lowered-text column, and the mask consumer counts only valid-and-true, so every blank was dropped. `criteria_match` itself was already correct (`Ne("Debt")` vs `Empty` returns true) — the non-cached fallback path was fine — but these formulas take the cached vectorized path, which is where the bug lived.

## Fix

In the `Ne(Text)` path, replace the `nilike`-produced NULLs (and the all-blank-segment NULL fast path) with the exact `<>`-vs-blank verdict, computed once from the pattern:

- `<>X` (non-empty pattern): a blank does not equal "X" → fill TRUE (blank counts)
- `<>` (empty pattern, i.e. "non-blank"): a blank equals "" → fill FALSE (blank does not count)

The per-cell fill only runs on chunks where `null_count() > 0`; blank-free chunks stay on the fully vectorized `nilike` path untouched, and an all-blank segment fills a constant. Numeric criteria, `Eq`, and wildcard paths are byte-identical to before.

## Finite range vs whole column

The important boundary: a blank satisfying `<>X` is correct for a finite range like `A1:A3`, but a whole-column `A:A` reference must not count the ~1M trailing empty cells. The fill is applied inside `iter_row_chunks`, which clamps iteration to the sheet's used rows, and `build_criteria_mask` returns an empty mask for views starting past the used region. So `<>` blank-inclusion is naturally scoped to the materialized region: `COUNTIFS(A1:A3,"<>Debt")` counts its interior blank, while `COUNTIFS(A:A,"<>Debt")` counts only used-region rows and never explodes. Both are pinned by tests.

## Behavior on a blank cell (pinned)

- `<>X` (non-empty): counts — this was the bug
- `<>` (non-blank): does not count
- `""` / `"="`: counts
- `"X"` / `"=X"`: does not count
- `>0`, `<5` (numeric): does not count (Excel doesn't count blanks for numeric comparisons — left unchanged)
- `ap*` (wildcard): does not count

## Tests

New `criteria_notequal_blank.rs` (7 tests): the exact #158 repro on finite ranges (SUMIFS=50, COUNTIFS=2, SUMIF=50, AVERAGEIFS=25), `<>` excluding blanks, empty-string/equality semantics, numeric predicates still excluding blanks, wildcard, multi-criteria with a blank-category row, and the whole-column non-explosion (bulk Arrow ingest → cached mask path → 96, matching the finite range, not ~1M). One existing test (`criteria_mask_text_chunked`) had a stale assertion that encoded the old all-NULL behavior for `<>` on an all-empty column; corrected to the definite all-false verdict.

fmt / clippy (`-D warnings`) / full workspace test suite all clean.

### DIFF
--- a/crates/formualizer-eval/src/engine/eval.rs
+++ b/crates/formualizer-eval/src/engine/eval.rs
@@ -1305,11 +1305,22 @@ fn compute_criteria_mask(
 
     // TEXT PATH: build masks per row-chunk using lowered text slices.
     // This avoids concatenating full-string columns just to compute a boolean mask.
+    //
+    // `ne_null_fill` is the value a blank/Empty cell must take under a `<>` criterion.
+    // Excel treats a blank as equal to the empty string, so:
+    //   - `<>X` (non-empty pattern): blank != "X" is TRUE  -> blank SATISFIES the criterion
+    //   - `<>`  (empty pattern):     blank == ""  is TRUE   -> blank does NOT satisfy `<>`
+    // Arrow's `nilike` returns NULL (not TRUE) for NULL inputs, so without this fill the
+    // vectorized path silently drops blanks for `<>X` (issue #158). This is bounded to the
+    // view's materialized rows because `iter_row_chunks`/`build_criteria_mask` only iterate
+    // the used region — a whole-column `A:A` ref never counts the millions of trailing blanks.
+    let mut ne_null_fill = false;
     let (text_kind, text_pat, empty_special) = match pred {
         crate::args::CriteriaPredicate::Eq(formualizer_common::LiteralValue::Text(t)) => {
             (0u8, t.to_lowercase(), t.is_empty())
         }
         crate::args::CriteriaPredicate::Ne(formualizer_common::LiteralValue::Text(t)) => {
+            ne_null_fill = !t.is_empty();
             (1u8, t.to_lowercase(), false)
         }
         crate::args::CriteriaPredicate::TextLike {
@@ -1353,6 +1364,13 @@ fn compute_criteria_mask(
                     let mut bb = BooleanBuilder::with_capacity(cs.row_len);
                     bb.append_n(cs.row_len, true);
                     bool_parts.push(bb.finish());
+                } else if text_kind == 1 {
+                    // `<>X`: an all-blank column segment is entirely blanks. Fill with the
+                    // Excel `<>`-vs-blank verdict (TRUE for `<>X`, FALSE for `<>`) instead of
+                    // dropping the rows to NULL (issue #158).
+                    let mut bb = BooleanBuilder::with_capacity(cs.row_len);
+                    bb.append_n(cs.row_len, ne_null_fill);
+                    bool_parts.push(bb.finish());
                 } else {
                     // For non-empty patterns, ilike/nilike return null on null inputs.
                     bool_parts.push(BooleanArray::new_null(cs.row_len));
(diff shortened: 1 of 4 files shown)


PSU3D0:
Closing in favor of #160 — @tansaku reported #158 and put up an equivalent fix (same root cause, the `nilike`-returns-null-on-blank issue), so it's right to land theirs. I'm porting the extra coverage from here onto their merge in a follow-up — the whole-column non-explosion test especially, plus the `null_count()` guard so blank-free `<>X` stays fully vectorized — so nothing here is lost.

PR #160, opened 2026-07-01
fix: SUMIFS/COUNTIFS/SUMIF <>text criteria should match blank cells (#158)

### LINKED ISSUE #158: SUMIFS/SUMIF/COUNTIFS "<>value" criteria exclude blank cells (Excel includes them)

tansaku:
## Summary

Fixes #158. A `<>value` criterion in `SUMIFS` / `SUMIF` / `COUNTIFS` now matches empty/blank cells, aligning with Excel and LibreOffice where a blank cell satisfies `<> "value"` (blank ≠ `"value"` is TRUE).

## Root cause

`compute_criteria_mask` in `crates/formualizer-eval/src/engine/eval.rs` builds not-equal text masks with `nilike`, which returns **null** for blank (null) input cells. Null mask entries are then treated as "no match" when the aggregation sums/counts, so blank-criteria rows were silently dropped from any `<>text` aggregation.

## Fix

For not-equal text predicates with a **non-empty** pattern, blanks are now treated as matching (`true`) - both for all-null segments and for individual null cells within a text segment. `<>""` (the "not blank" case) is deliberately left unchanged, so a blank still does not match `<>""`.

Diff is 3 lines in `eval.rs`.

## Tests

Added `crates/formualizer-eval/src/engine/tests/sumifs_ne_blank_158.rs` reproducing the issue (Debt / blank / Equity rows) through the Arrow evaluation path:

| Case | Before | After (expected, Excel/LibreOffice) |
|---|---|---|
| `SUMIFS(B, A, "<>Debt")` | 30 | 50 |
| `COUNTIFS(A, "<>Debt")` | 1 | 2 |
| `SUMIF(A, "<>Debt", B)` | 30 | 50 |

`cargo fmt --all --check`, `cargo clippy -p formualizer-eval --all-targets -- -D warnings`, and the `formualizer-eval` lib tests all pass (the pre-existing `IMEXP`/`IMSIN`/`IMCOS` ULP test failures are unrelated to this change).

---

Prepared with AI assistance and reviewed by the author.

### DIFF
--- a/crates/formualizer-eval/src/engine/eval.rs
+++ b/crates/formualizer-eval/src/engine/eval.rs
@@ -1326,6 +1326,7 @@ fn compute_criteria_mask(
         _ => return None,
     };
 
+    let ne_matches_blank = text_kind == 1 && !text_pat.is_empty();
     let pat = StringArray::new_scalar(text_pat);
     let mut bool_parts: Vec<BooleanArray> = Vec::new();
 
@@ -1348,7 +1349,7 @@ fn compute_criteria_mask(
             None => {
                 #[cfg(test)]
                 criteria_mask_test_hooks::inc_all_null();
-                if text_kind == 0 && empty_special {
+                if (text_kind == 0 && empty_special) || ne_matches_blank {
                     // Eq("") treats nulls (Empty) as equal.
                     let mut bb = BooleanBuilder::with_capacity(cs.row_len);
                     bb.append_n(cs.row_len, true);
@@ -1369,7 +1370,7 @@ fn compute_criteria_mask(
             _ => return None,
         };
 
-        if text_kind == 0 && empty_special {
+        if (text_kind == 0 && empty_special) || ne_matches_blank {
             // Treat nulls as equal to empty string
             let mut bb = BooleanBuilder::with_capacity(seg_sa.len());
             for i in 0..seg_sa.len() {
--- a/crates/formualizer-eval/src/engine/tests/mod.rs
+++ b/crates/formualizer-eval/src/engine/tests/mod.rs
@@ -135,6 +135,7 @@ mod sumif_arrow_used_bounds;
 mod sumifs_arrow_edits;
 mod sumifs_arrow_fastpath;
 mod sumifs_cached_mask_padding;
+mod sumifs_ne_blank_158;
 mod used_bounds_cache;
 mod whole_column_sumifs;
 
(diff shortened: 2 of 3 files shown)


PSU3D0 (approved):
Thanks Sam — clean fix, and it matches exactly where I'd traced the root cause (`nilike` returning null for blank inputs on `<>`). The `or_kleene` with `is_null` is a tidy way to fold the blanks in. Merging.

I'll follow up with a bit more coverage on top — mainly a whole-column `A:A` case to lock in that `<>X` stays bounded to the used region and never starts counting the trailing blanks, plus a couple of edge cases (`<>` excluding blanks, `>0` still excluding them) and a small guard so blank-free `<>X` chunks skip the null pass entirely. Nice to have the reporter fix their own bug.

PR #197, opened 2026-07-19
Render #REF sheet-sentinel references as the #REF! literal

phrrngtn:
HI,
I am not a Rust programmer but I was doing some experimentation with formualizer via Claude Code and noticed some discrepancies wrt some Python-based Excel formula-parsing. Most of these appear to be fixed on main: I  was going against the published crate and encountered more errors but all but one went away when I tested against main. The one remaining one is described in the paragraph below, authored by Claude code, which also wrote the fix and the test-cases. So I hope that even if the fix is not appropriate, then at least the problem is described in sufficient detail. I am doing some experimentation with Office.js addins (most of my experience is with PyXLL (Python) and ExcelDNA(C#)) and wanted a formula parser that could be compiled to WASM easily and this is where I encountered your library for the first time. I am looking forward to working with it more on as part of an addin that does Excel formula "linting".

A reference invalidated by a structural delete carries a Cell/Range with sheet == "#REF". ReferenceType's Display did not special-case this, so it rendered as a quoted sheet plus a bogus address (e.g. '#REF'!0, or '#REF'!$$0 when the original had absolute flags). Render it as the Excel #REF! literal instead, in both the Cell and Range arms. Adds a regression test.

### DIFF
--- a/crates/formualizer-parse/src/parser.rs
+++ b/crates/formualizer-parse/src/parser.rs
@@ -1008,6 +1008,13 @@ impl Display for ReferenceType {
                     row_abs,
                     col_abs,
                 } => {
+                    // A reference invalidated by a structural delete carries the
+                    // "#REF" sheet sentinel; render it as the Excel #REF! literal
+                    // rather than a quoted sheet with a bogus address.
+                    if sheet.as_deref() == Some("#REF") {
+                        return write!(f, "#REF!");
+                    }
+
                     let col_str = Self::format_col(*col, *col_abs);
                     let row_str = Self::format_row(*row, *row_abs);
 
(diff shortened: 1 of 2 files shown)


PSU3D0:
Thank you for finding and isolating this. The malformed `'#REF'!0` output is a real bug, and your regression report led us to a deeper representation issue.

We cannot merge the sheet-name display special case unchanged because `#REF` is also a valid worksheet name, so `='#REF'!A1` would be incorrectly rendered as `#REF!`. The existing sentinel also bypasses normal dependency rewiring and incorrectly poisons lazy formulas such as `IF(FALSE, deleted_ref, 5)` and recoverable formulas such as `IFERROR(deleted_ref, 5)`.

I opened #206 to supersede this with a canonical fix: structural deletes now rewrite the affected AST leaf to the same `Literal(Error(Ref))` representation produced by parsing `#REF!`. It also covers sheet locality, legitimate `#REF` sheets, whole-axis ranges, undo/redo, names, and FormulaPlane paths. Your commit is credited as a co-author.

I will leave this PR open until #206 lands, then close it as superseded. Thanks again for the clear report and starting point.

PSU3D0:
Superseded by #206, which is now merged as 71e08bd0. Thank you again for the report and starting point; the merged commit includes co-author credit.

PR #202, opened 2026-07-19
tokenizer: treat CR and TAB as whitespace (fix CRLF parse failures)

phrrngtn:
Hi, 
this is another minor bug report and fix, this time for embedded newline  and carriage return characters in formulas. I found it while doing some back-testing on parsing formulas from large corpuses of spreadsheets. Claude Code generated the fix, the test fixtures, and the remainder of this text in the PR. All existing tests continued to pass with the fix incorporated so I hope you will the PR fit for use. I guess for completeness you could put in the two remaining ASCII whitespace characters, \v and \f (vertical tab and form-feed, respectively) but I have never encountered them in the wild as whitespace.

The tokenizer's whitespace set was `b' ' | b'\n'` — only space and LF. Carriage return (`\r`) and tab (`\t`) were not whitespace, so they fell through to the operand accumulator. A lone `\r` (e.g. in `SUM(A1,\rA2)`) usually got absorbed into an adjacent operand and survived, but the Windows CRLF sequence `\r\n` produced a stray operand token consisting of just `\r`, which the parser then rejected — so multi-line formulas entered with Alt+Enter (`SUM(\r\n1,2\r\n)`) failed to parse even though `\r` and `\n` individually did not.

Add `\r` and `\t` to every whitespace site (both the Tokenizer and SpanTokenizer dispatch + parse_whitespace consume loops, the boundary check, and the token-run scanner), mirroring the existing `\n` handling. parse_whitespace already saves the pending token, so no TOKEN_ENDERS change is needed.

Found by stress-testing the parser over a large real corpus (Enron) and independently reproduced by Apache POI's own `SUM(\r\n1,2\r\n)` test-data case.

### DIFF
--- a/crates/formualizer-parse/src/tests/tokenizer.rs
+++ b/crates/formualizer-parse/src/tests/tokenizer.rs
@@ -2138,5 +2138,27 @@ mod tests {
                         .any(|t| t.0 == TokenType::OpInfix && t.2 == " ")
             );
         }
+
+        #[test]
+        fn crlf_and_tabs_are_whitespace() {
+            // Regression: the tokenizer treated only ' ' and '\n' as whitespace,
+            // so the Windows CRLF sequence "\r\n" (Alt+Enter line breaks) and tabs
+            // broke parsing — even though '\r' and '\n' individually did not.
+            for f in [
+                "=SUM(\r\n1,2\r\n)", // Apache POI test-data case
+                "=A1 +\r\n B1",
+                "=IF(A1,\r\nB1)",
+                "=SUM(A1,\tA2)",
+                "=\tSUM(A1,A2)\t",
+            ] {
+                assert!(Tokenizer::new(f).is_ok(), "failed to tokenize {f:?}");
+            }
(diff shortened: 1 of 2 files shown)


PSU3D0:
Thank you for finding this. The CRLF/TAB diagnosis and the boundary updates are correct, and I have preserved your original commit in #207.

The additional change in #207 handles one interaction with the parser's space-intersection operator. TAB, CR, and LF are accepted as ordinary lexical whitespace, but only an actual ASCII space in the run can enable range intersection. When intersection is present, owned tokens and the AST use the canonical operator value `" "`, while source-backed spans and rendering preserve the exact original CR/LF/TAB bytes.

The replacement branch adds exact parity tests for both tokenizer frontends, parser ASTs, source spans, quoted strings/sheets, CR/LF/TAB-only reference boundaries, and mixed whitespace intersections. Your contribution is credited as co-author.

I will leave this PR open until #207 lands, then close it as superseded. Thanks again for the useful corpus-derived report.

PSU3D0:
Superseded by #207, now merged. Thank you again for the corpus-derived report and fix; the merged change includes co-author credit.

PR #213, opened 2026-07-21
fix(eval): expand range arguments in CONCAT and TEXTJOIN

### LINKED ISSUE #209: [bug] `TEXTJOIN` and `CONCAT` silently ignore all but the first cell of a range argument

## Summary

`TEXTJOIN` and `CONCAT` accept a range argument without error, but only the first cell of the range reaches the result. The rest are dropped silently — no error value, no warning.

Literal arguments and individual cell arguments are handled correctly, and `SUM` over the same shape of range is fine, so this looks specific to range expansion in the text functions rather than to range resolution generally.

The silent part is what makes it costly: the formula looks right and returns a plausible string, so the truncation is easy to miss.

## Reproduction

```js

await initializeWasm();
const wb = new Workbook();
const S = 'Sheet1';

// A2:A5 = names, C2:C5 = numbers
[['Ani', 10], ['Budi', 20], ['Cici', 30], ['Dodi', 40]].forEach(([name, n], i) => {
  wb.setValue(S, i + 2, 1, name);
  wb.setValue(S, i + 2, 3, n);
});

const cases = [
  '=TEXTJOIN("-",TRUE,"a","b","c")', // literal args
  '=TEXTJOIN("-",TRUE,A2,A3)',       // individual cell args
  '=TEXTJOIN("-",TRUE,A2:A5)',       // range arg
  '=CONCAT("a","b","c")',
  '=CONCAT(A2:A5)',
  '=SUM(C2:C5)',                     // control: ranges work elsewhere
];

cases.forEach((f, i) => wb.setFormula(S, 10 + i * 2, 6, f));
wb.evaluateAll();
cases.forEach((f, i) => console.log(f.padEnd(34), '->', JSON.stringify(wb.evaluateCell(S, 10 + i * 2, 6))));
```

## Actual

```
=TEXTJOIN("-",TRUE,"a","b","c")    -> "a-b-c"
=TEXTJOIN("-",TRUE,A2,A3)          -> "Ani-Budi"
=TEXTJOIN("-",TRUE,A2:A5)          -> "Ani"      <-- only the first cell
=CONCAT("a","b","c")               -> "abc"
=CONCAT(A2:A5)                     -> "Ani"      <-- only the first cell
=SUM(C2:C5)                        -> 100        (control, correct)
```

## Expected (Excel)

```
=TEXTJOIN("-",TRUE,A2:A5)          -> "Ani-Budi-Cici-Dodi"
=CONCAT(A2:A5)                     -> "AniBudiCiciDodi"
```

## Confirmed on a second, unrelated fixture

Different data, different columns (E:H, rows 3-12), and different range shapes — same behaviour throughout:

```
=TEXTJOIN("|",TRUE,E3:E8)          -> "Pensil"        vertical range
=TEXTJOIN("|",TRUE,E12:H12)        -> "a"             horizontal range
=TEXTJOIN("|",TRUE,E3:F4)          -> "Pensil"        2-D range
=TEXTJOIN(",",TRUE,G3:G5)          -> "2500"          numeric range
=TEXTJOIN("|",FALSE,E3:E5)         -> "Pensil"        ignore_empty = FALSE
=CONCAT(E3:E5)                     -> "PensilBuku…"?  no: "Pensil"
=CONCAT(E12:H12)                   -> "a"
=SUM(G3:G8)                        -> 1400000         control, correct
=COUNTA(E3:E8)                     -> 6               control, correct
```

Two cases that pin the behaviour down more precisely than "the first argument wins":

```
=TEXTJOIN("|",TRUE,E3:E4,E12:F12)  -> "Pensil|a"      each range contributes exactly one cell
=TEXTJOIN("|",TRUE,E3:E4,"zz")     -> "Pensil|zz"     literals alongside ranges still fine
```

So every range argument is collapsed to its top-left cell, independently, wherever it appears in the argument list. Orientation, dimensionality and data type make no difference, and `ignore_empty` is not involved.

## Note

`CONCATENATE(A2:A5)` also returns `"Ani"`, but I am deliberately leaving it out of this report: Excel's legacy `CONCATENATE` does not expand ranges the way `CONCAT` does, so that result may well be intentional.

maulafaizz:
Fixes #209.

## Problem

`CONCAT` and `TEXTJOIN` read their value arguments through `scalar_like_value`, which resolves a range to `rv.get_cell(0, 0)`. Every range argument was therefore reduced to its top-left cell, with no error raised:

```
=TEXTJOIN("-",TRUE,A2:A5)          -> "Ani"        want "Ani-Budi-Cici-Dodi"
=CONCAT(A2:A5)                     -> "Ani"        want "AniBudiCiciDodi"
=TEXTJOIN("|",TRUE,A2:A3,C1:D1)    -> "Ani|x"      each range collapsed independently
```

Literal and single-cell arguments were unaffected, and `SUM` over the same ranges was correct, so this was isolated to the text functions' argument handling.

## Change

Adds `value_parts`, which expands an argument into the values it contributes — a scalar yields one value, a range yields every cell in row-major order via `for_each_cell` (the same idiom already used in `array_text.rs`) — and uses it for the value arguments of `CONCAT` and `TEXTJOIN`.

`TEXTJOIN` keeps its existing per-value coercion and `ignore_empty` handling, which now also applies to blank cells inside a range. The text coercion body is split out of `to_text` into `literal_to_text` so the scalar and expanded paths share it.

No `arg_schema` changes, so the generated schema blocks are untouched.

## `CONCATENATE` deliberately not changed

`CONCATENATE(A2:A5)` still returns `"Ani"`. Excel's legacy `CONCATENATE` does not expand ranges the way `CONCAT` does, so that behaviour may well be intended — happy to change it if you'd rather the two stay identical, since the doc comment currently says they share semantics.

## Tests

Four added, covering vertical, horizontal and multi-range expansion, literals interleaved with ranges, and `ignore_empty = FALSE` keeping blanks inside a range.

## Verification

Run with Rust 1.96.1 (nixpkgs), since I didn't have the mise-pinned 1.93 available:

- `cargo test -p formualizer-eval` — 2395 passed, 1 failed. The failure is `run_formula_test_suite` (914 passed / 16 skipped / 3 failed), and it fails identically on unmodified `main`: the three cases are complex-number formatting (`-13.128783081462158-15.200784463067954i` and two similar), unrelated to this change.
- `cargo fmt --all` — clean.
- `cargo clippy -p formualizer-eval --all-targets` — no warnings in the touched file. Note the crate does not currently pass `-D warnings` under clippy 1.96 on `main` either (22 errors before this change), which I assume is just the newer toolchain.
- `cargo run -p xtask -- schema` and `-- audit --strict` — no stale entries, no issues, no files modified. Both report scanning nothing (`files scanned: 0`, `total registered: 0`), so treat that line as a no-op rather than as coverage.

Small note: CONTRIBUTING.md refers to `xtask -- docs-audit` / `docs-schema`, but the subcommands are now `audit` / `schema`.


PSU3D0:
Thank you for this entire batch of reports and PRs. The reproductions are unusually strong: they exercise the published WASM surface, include useful controls, and isolate contradictions that our existing builtin tests missed. I confirmed that the reported problems are real, and these PRs give us excellent regression signal and practical starting points.

Across #213, #214, #215, #217, and #218, the remaining work is mostly maintainer hardening around shared contracts and edge behavior rather than doubt about the underlying reports. We plan to start those passes later this week. There is no pressure to complete that hardening yourself; if you would like to take another pass at any item, leave a note on that PR so we can avoid duplicating work.

For this PR specifically, the plan is to keep legacy `CONCATENATE` non-expanding, use one streaming expansion path for scalar/range/computed-array values, align argument schemas and generated metadata with runtime behavior, and add focused 2-D, computed-array, error, and blank-value coverage.

PSU3D0:
Rebased onto current `main` and landed with maintainer hardening on top of your commit.

What was added beyond the original change:

- Range and array expansion now goes through the shared `resolve_once` argument path introduced in #217, so `CONCAT`/`TEXTJOIN` resolve computed and reference arguments exactly once rather than re-evaluating them.
- Legacy `CONCATENATE` keeps its top-left-only semantics instead of aliasing `CONCAT`, which is the actual Excel behavior and was previously undocumented.
- Excel's 32,767-character result limit is enforced, with boundary tests on both sides.
- Error propagation order through expanded ranges is pinned, including an error in a non-first cell.

One nuance worth recording: only the error *kind* survives array expansion, not its diagnostic message, because materializing an array argument goes through the Arrow-backed range view. That is engine-wide behavior since #217 rather than anything specific to these two functions, so the test asserts the kind.

Your commit and authorship are preserved. Closes #209.

PR #214, opened 2026-07-21
fix(eval): use the shared Excel serial mapping in TEXT date formatting

### LINKED ISSUE #210: [bug] `TEXT` shifts every date after serial 59 by one day (does not emulate Excel's phantom 1900-02-29)

## Summary

`TEXT` renders date serials one day later than Excel for every serial above 59, while `DAY`, `MONTH` and `DATE` on the same serials are correct. The two paths disagree about the same number.

The cause looks like Excel's well-known 1900 leap-year bug: Excel treats serial 60 as the non-existent 1900-02-29, so serial 61 is 1900-03-01. `DATE`/`DAY`/`MONTH` here reproduce that behaviour, but `TEXT` appears to convert serials with a real proleptic calendar, skipping the phantom day and shifting everything after it forward by one.

Not timezone-related — identical output under `TZ=UTC`, `TZ=Asia/Jakarta` and `TZ=America/New_York`.

## Reproduction

```js

await initializeWasm();
const wb = new Workbook();
const S = 'Sheet1';

const cases = [
  '=TEXT(1,"yyyy-mm-dd")',      // Excel: 1900-01-01
  '=TEXT(59,"yyyy-mm-dd")',     // Excel: 1900-02-28
  '=TEXT(61,"yyyy-mm-dd")',     // Excel: 1900-03-01
  '=TEXT(45306,"yyyy-mm-dd")',  // Excel: 2024-01-15
  '=DAY(45306)',                // Excel: 15
  '=MONTH(45306)',              // Excel: 1
  '=DATE(2024,1,15)',           // Excel: 45306
  '=TEXT(1234.5,"0.00")',       // control: numeric formatting
];

cases.forEach((f, i) => wb.setFormula(S, 10 + i * 2, 6, f));
wb.evaluateAll();
cases.forEach((f, i) => console.log(f.padEnd(30), '->', JSON.stringify(wb.evaluateCell(S, 10 + i * 2, 6))));
```

## Actual

```
=TEXT(1,"yyyy-mm-dd")          -> "1900-01-01"   correct
=TEXT(59,"yyyy-mm-dd")         -> "1900-02-28"   correct
=TEXT(61,"yyyy-mm-dd")         -> "1900-03-02"   <-- Excel: 1900-03-01
=TEXT(45306,"yyyy-mm-dd")      -> "2024-01-16"   <-- Excel: 2024-01-15
=DAY(45306)                    -> 15             correct
=MONTH(45306)                  -> 1              correct
=DATE(2024,1,15)               -> 45306          correct
=TEXT(1234.5,"0.00")           -> "1234.50"      correct
```

The boundary is clean: serials ≤ 59 render correctly, serials ≥ 61 are one day late.

## Expected (Excel)

```
=TEXT(61,"yyyy-mm-dd")         -> "1900-03-01"
=TEXT(45306,"yyyy-mm-dd")      -> "2024-01-15"
```

## Confirmed on further serials and formats

```
=TEXT(58,"yyyy-mm-dd")             -> "1900-02-27"   correct
=TEXT(60,"yyyy-mm-dd")             -> "1900-03-01"   <-- Excel: 1900-02-29 (the phantom day)
=TEXT(100,"yyyy-mm-dd")            -> "1900-04-10"   <-- Excel: 1900-04-09
=TEXT(1000,"yyyy-mm-dd")           -> "1902-09-27"   <-- Excel: 1902-09-26
=TEXT(40000,"yyyy-mm-dd")          -> "2009-07-07"   <-- Excel: 2009-07-06
=TEXT(45306,"dd/mm/yyyy")          -> "16/01/2024"   <-- Excel: 15/01/2024
=TEXT(I30,"yyyy-mm-dd")            -> "2024-01-16"   <-- from a cell holding 45306
=YEAR(40000)                       -> 2009           correct
=DAY(40000)                        -> 6              correct
```

Serial 60 is the clearest case: Excel renders it as the non-existent `1900-02-29`, this renders `1900-03-01`. The format string makes no difference, and the same shift occurs whether the serial is a literal or read from a cell.

Round-tripping drifts, too:

```
=DATEVALUE(TEXT(45306,"yyyy-mm-dd"))   -> 45307   <-- Excel: 45306
```

so `DATEVALUE` and `TEXT` are consistent with each other but both diverge from `DATE`/`DAY`/`MONTH`/`YEAR`.

## Why it matters

`DATE(2024,1,15)` and `TEXT(DATE(2024,1,15),"yyyy-mm-dd")` disagree inside the same workbook, so any formula that formats a date it just computed produces a wrong string. Numeric formats (`"0.00"`) are unaffected, so this is isolated to the date branch of `TEXT`'s serial decoding.

maulafaizz:
Fixes #210.

## Problem

`format_serial_date` in `builtins/text/value_text.rs` added the truncated serial to `1899-12-31` directly. The comment above it already flagged this — *"very naive: treat integer part as days since 1899-12-31 ignoring leap bug for now"*.

Excel counts a phantom `1900-02-29` at serial 60, so skipping it shifts every date from serial 61 onward one day late:

```
=TEXT(61,"yyyy-mm-dd")     -> 1900-03-02   Excel: 1900-03-01
=TEXT(100,"yyyy-mm-dd")    -> 1900-04-10   Excel: 1900-04-09
=TEXT(45306,"yyyy-mm-dd")  -> 2024-01-16   Excel: 2024-01-15
```

`DATE`, `DAY`, `MONTH` and `YEAR` were already right, because they go through `datetime::serial::serial_to_date`. So a workbook disagreed with itself: `DATE(2024,1,15)` gives `45306`, and formatting that same serial gave `2024-01-16`. Round-tripping drifted too — `DATEVALUE(TEXT(45306,"yyyy-mm-dd"))` returned `45307`.

## Change

Route `format_serial_date` through the existing `serial_to_date` rather than keeping a second, divergent mapping. No module boundaries move; `datetime/mod.rs` already re-exports it.

That does introduce an error path the old arithmetic never had — `serial_to_date` rejects negative serials and serials past the end of the calendar, where the old code simply kept adding days. Those now produce `#VALUE!`, which is how this function already reports input it cannot format.

To be explicit, that second part is a deliberate behaviour change rather than a preserved one: `TEXT(-1,"yyyy-mm-dd")` used to render `1899-12-30`. Happy to render something on that path instead if you would rather it stay total.

## One case this does not change

`TEXT(60,...)` now renders `1900-02-28`, where Excel shows `1900-02-29`.

That follows `serial_to_date`'s deliberate choice, documented in `serial.rs`: *"we do NOT invent an impossible date object"*. Making TEXT print `1900-02-29` would need a display-level special case and would then disagree with `DAY(60)`, which returns 28 today. Consistency inside the engine seemed the better default, but say the word if you'd rather the display path emulate the phantom day exactly.

## Tests

Two added:

- a case table covering the boundary (serials 1, 59, 61), the modern serials from the issue, and a second format string to confirm the fix is not tied to `yyyy-mm-dd`;
- a negative serial with a date format, pinning the `#VALUE!` described above.

## Verification

Rust 1.96.1 (nixpkgs), as I didn't have the mise-pinned 1.93 available:

- `cargo test -p formualizer-eval` — 2393 passed, 1 failed. The failure is `run_formula_test_suite` at `914 passed / 16 skipped / 3 failed`, identical to unmodified `main`; the three cases are complex-number formatting, unrelated to dates.
- `cargo fmt --all` — clean.
- `cargo clippy -p formualizer-eval --all-targets` — no warnings in either touched file. (The crate does not pass `-D warnings` under clippy 1.96 on `main` either, which I take to be toolchain drift from the pinned 1.93.)
- `cargo run -p xtask -- schema` / `-- audit --strict` — clean and no files modified, though both report `files scanned: 0` / `total registered: 0` for me, so I would treat that as a no-op rather than as real coverage. Let me know if they need a flag I have missed.

Stacked on the same fork as #213 but touches different files, so the two are independent.


maulafaizz:
Force-pushed a simplification: the first version changed `mod serial;` to `pub(crate) mod serial;`, which was unnecessary — `datetime/mod.rs` already does `pub use serial::*;`, so `crate::builtins::datetime::serial_to_date` was reachable all along. Dropped the visibility change; the diff is now two files and touches no module boundaries. Same tests, same results.

maulafaizz:
Added one commit (`c7261f0`) rather than force-pushing, since it changes behaviour rather than tidying the same change.

Routing through `serial_to_date` introduced an error path the old arithmetic never had — negative serials, and serials past the end of the calendar. The first version returned the format string verbatim on that path, so `TEXT(-1,"yyyy-mm-dd")` rendered as the literal `"yyyy-mm-dd"`, and I described it in the PR body as preserving the old behaviour, which it did not: the old code produced `1899-12-30` there.

It now returns `#VALUE!`, matching how `TEXT` already reports input it cannot format, with a test pinning it. The body is corrected. Still happy to make that path render a date instead if you would rather it stay total.

PSU3D0:
Confirmed: the shared serial conversion fixes the post-60 offset for ordinary 1900-system dates. Before merge, we plan to complete the boundary behavior for the 1904 date system, display-only serial 60 compatibility, and invalid/non-finite/out-of-calendar serials, with formula-level boundary and round-trip tests.

The broader coordination note is on #213. If you would like to take another pass at these cases before our maintainer pass later this week, leave a note here; otherwise, no action is required.

PR #215, opened 2026-07-21
docs: correct xtask subcommand names in CONTRIBUTING

maulafaizz:
The builtin-docs commands in CONTRIBUTING.md fail as written:

```
$ cargo run -p xtask -- docs-schema
error: unrecognized subcommand 'docs-schema'
```

`cargo run -p xtask -- --help` lists the subcommands as `audit`, `schema` and `ref`. Renamed the four occurrences accordingly.

The flags are unchanged and still correct as documented — `audit` takes `--strict`, `schema` takes `--apply` and `--allow-dirty` — so only the subcommand names move.

I left the two mentions in `docs/architecture/builtin-docs-structured-metadata.md` alone ("Lint checks to add in `docs-audit`"), since those read as design prose about future work rather than commands a contributor would copy. Happy to update them too if you'd rather the naming be consistent everywhere.

Noticed this while running the pre-PR checks for #213.


PSU3D0:
Confirmed: the `CONTRIBUTING.md` replacements match the actual `audit`, `schema`, and `ref` CLI. We plan to complete the remaining copyable reference in `.gitattributes` and separately repair xtask discovery, which still searches for `register_function` after builtin registration moved to `register_builtin` and therefore reports a false-success zero inventory.

The broader coordination note is on #213. You are welcome to take a pass at either item before our maintainer work later this week, but there is no expectation that this documentation PR absorb the scanner repair.

PR #199, opened 2026-07-19
feat(workbook): prototype calculation packages

PSU3D0:
## Spike scope

This draft PR proves only S0/S1 of the post-C5 calculation-artifact exploration. It does not implement artifact-closure pruning, claim a stable production format, or serialize `RecalcPlan` or private engine identities.

## S0: bounded capability discovery

- independently scan ZIP/OOXML source facts before lossy adapter conversion
- enforce entry/path/size/ratio/manifest admission limits and relationship-resolved allow-listing
- report date-system, names, formula-cache presence, tables, hidden state, active parts, external links, macros, and backend fidelity
- compute deterministic source and canonical manifest digests
- refuse unsupported, lossy, malformed, incomplete, or cancelled discovery with typed reasons

## S1: deterministic whole-workbook package

- feature-gated experimental fixed-header canonical JSON container
- preserve the proven modeled subset of stored values/errors, canonical formulas at original coordinates, sheet order/dimensions, supported names/tables, row visibility, date/cycle settings, manifest metadata, and digests
- rebuild formulas after load; cached XLSX formula results are intentionally excluded
- reject computed-overlay compaction so calculated results cannot be laundered into stored source values
- validate loader formulas, volatility, dimensions, names, tables, coordinates, and binding expectations transactionally

## Proven subset and limits

The spike proves deterministic in-process builds and real Umya XLSX S0 → digest-bound S1 → load/evaluate/rebuild parity for the supported subset. Arrow currently normalizes some integer/date/time distinctions, cached results are excluded, Calamine table loss is refused, and fresh-process determinism remains a later oracle.

S2 still requires a new resource-accounted stable artifact-closure query. Existing target-preparation reports are not export closures.

## Validation

- 9 focused experimental FZCP tests, including real Umya XLSX parity
- default and no-default experimental checks for workbook/eval
- strict all-target Clippy for experimental features and Umya-backed tests
- `cargo fmt --all -- --check`
- `git diff --check`

Fresh architecture review returned `SPIKE_VALID: YES` and `S2_READY: YES`; the final Calamine table-loss and resource-hardening findings were fixed afterward.

PR #217, opened 2026-07-21
fix(eval): let aggregates consume computed array results

### LINKED ISSUE #211: [bug] Dynamic-array results spill correctly but cannot be consumed by aggregate functions

## Summary

`UNIQUE`, `SEQUENCE` and `SORT` spill correctly when written on their own. But passing one of them *as an argument* to an aggregate function fails: `SUM` returns `#VALUE!`, and `COUNTA` silently counts 1 instead of the array length.

The array clearly exists and has the right shape — `ROWS(UNIQUE(B2:B5))` returns 3, and `INDEX(SORT(...),1)` returns the right element. So shape-inspecting and element-selecting functions see the array while aggregating functions collapse it to a single value.

`COUNTA` is the concerning case: it returns a plausible number rather than an error, so the mistake is silent.

## Reproduction

```js

await initializeWasm();
const wb = new Workbook();
const S = 'Sheet1';

[['Ani', 'Jakarta', 10], ['Budi', 'Bandung', 20], ['Cici', 'Jakarta', 30], ['Dodi', 'Surabaya', 40]]
  .forEach((row, i) => row.forEach((v, c) => wb.setValue(S, i + 2, c + 1, v)));

// 1. spilling on its own -- one array formula per column, nothing below it
wb.setFormula(S, 10, 6, '=UNIQUE(B2:B5)');
wb.setFormula(S, 10, 8, '=SEQUENCE(4)');
wb.setFormula(S, 10, 10, '=SORT(C2:C5,1,-1)');
wb.evaluateAll();
for (const [name, col] of [['UNIQUE', 6], ['SEQUENCE', 8], ['SORT', 10]]) {
  console.log('  ' + name.padEnd(10), JSON.stringify([0, 1, 2, 3].map((d) => wb.evaluateCell(S, 10 + d, col))));
}

// 2. the same arrays used as arguments
const cases = [
  '=ROWS(UNIQUE(B2:B5))',                // Excel: 3
  '=INDEX(SORT(C2:C5,1,-1),1)',          // Excel: 40
  '=COUNTA(UNIQUE(B2:B5))',              // Excel: 3
  '=SUM(SEQUENCE(4))',                   // Excel: 10
  '=SUM(SORT(C2:C5,1,-1))',              // Excel: 100
];
cases.forEach((f, i) => wb.setFormula(S, 30 + i * 2, 12, f));
wb.evaluateAll();
cases.forEach((f, i) => console.log('  ' + f.padEnd(40), '->', JSON.stringify(wb.evaluateCell(S, 30 + i * 2, 12))));
```

## Actual

```
spilling on its own -- all correct:
  UNIQUE     ["Jakarta","Bandung","Surabaya",null]
  SEQUENCE   [1,2,3,4]
  SORT       [40,30,20,10]

used as an argument:
  =ROWS(UNIQUE(B2:B5))                  -> 3           correct
  =INDEX(SORT(C2:C5,1,-1),1)            -> 40          correct
  =COUNTA(UNIQUE(B2:B5))                -> 1           <-- Excel: 3 (silent)
  =SUM(SEQUENCE(4))                     -> "#VALUE!"   <-- Excel: 10
  =SUM(SORT(C2:C5,1,-1))                -> "#VALUE!"   <-- Excel: 100
```

## Expected (Excel)

```
=COUNTA(UNIQUE(B2:B5))                -> 3
=SUM(SEQUENCE(4))                     -> 10
=SUM(SORT(C2:C5,1,-1))                -> 100
```

## Confirmed on a second, unrelated fixture and more aggregates

Different data and columns, plus `TRANSPOSE` as a fourth array producer:

```
=AVERAGE(SEQUENCE(4))              -> "#DIV/0!"   <-- Excel: 2.5
=COUNT(SEQUENCE(5))                -> 0           <-- Excel: 5
=MAX(SORT(G3:G8,1,-1))             -> 0           <-- Excel: 800000
=SUM(UNIQUE(G3:G8))                -> "#VALUE!"   <-- Excel: 1400000
=COUNTA(UNIQUE(F3:F8))             -> 1           <-- Excel: 3
=SUM(TRANSPOSE(G3:G5))             -> "#VALUE!"   <-- Excel: 467500
=ROWS(UNIQUE(F3:F8))               -> 3           correct
=INDEX(UNIQUE(F3:F8),2)            -> "Furnitur"  correct
=SUM(G3:G5)                        -> 467500      control, correct
```

`COUNT` returning `0` and `AVERAGE` returning `#DIV/0!` are the most telling: the aggregate is not merely mis-sizing the array, it is seeing **no numeric values at all**. `MAX` returning `0` rather than an error fits the same picture. Meanwhile `ROWS` and `INDEX` read the same expression correctly, so the array is present and correctly shaped — it just does not survive being fed to a function that iterates values.

## Note

An earlier revision of this report also listed `SUM(FILTER(C2:C5,B2:B5="Jakarta"))` here. That was a mistake on my part: `FILTER` fails earlier and for an unrelated reason, and it is tracked separately in #216. I have removed it so this issue describes a single cause; the correction is also recorded in the comment below.

maulafaizz:
Fixes #211.

## Problem

`ArgumentHandle::range_view()` resolves a function-call or operator argument only through `reference_for_eval()`. That works for functions returning a *reference* (`OFFSET`, `INDIRECT`), but not for expressions that produce a computed array.

Every aggregate gates on `if let Ok(view) = arg.range_view()`, so those arrays never arrived and the aggregate silently fell through to a path that saw no numeric values:

```
=SUM(SEQUENCE(4))          -> #VALUE!    Excel: 10
=COUNT(SEQUENCE(5))        -> 0          Excel: 5
=AVERAGE(SEQUENCE(4))      -> #DIV/0!    Excel: 2.5
=COUNTA(UNIQUE(B2:B5))     -> 1          Excel: 3
=MAX(SORT(C2:C5,1,-1))     -> 0          Excel: 800000
=SUM(TRANSPOSE(C2:C5))     -> #VALUE!    Excel: 467500
=SUM((B2:B5="Jakarta")*1)  -> #VALUE!    Excel: 2
```

`COUNT` returning `0` and `AVERAGE` returning `#DIV/0!` are the tell: the aggregate wasn't mis-sizing the array, it saw nothing at all. Meanwhile `ROWS(UNIQUE(...))` and `INDEX(SORT(...),1)` were correct, because they consume the value rather than asking for a range view.

## Change

Adds `range_view_from_computed_value()`, used as a fallback when `reference_for_eval()` fails. It materialises the evaluated argument into a `RangeView` via `from_owned_rows` — the same construction the `ASTNodeType::Array` arm already uses — covering both forms a computed array arrives in:

- `CalcValue::Range` — produced by `SEQUENCE`, `UNIQUE`, `SORT`, `TRANSPOSE`
- `CalcValue::Scalar(LiteralValue::Array)` — produced by a range comparison or elementwise operation, e.g. `B2:B5="Jakarta"` or `(B2:B5="x")*1`

Applied to both the AST and arena branches, which carried the identical pattern.

A genuine non-array scalar deliberately still errors rather than widening to a 1×1 range, since that would change the meaning of existing formulas.

## Tests

Two added in `aggregate.rs`, one per form: `SUM` over `SORT(A1:A3)` (asserting 60), and `SUM` over `(A1:A3=10)*1` (asserting 2). I verified each fails on unmodified `main` and passes with the change, so they pin the behaviour rather than merely passing.

## Verification

Rust 1.96.1 (nixpkgs), as I didn't have the mise-pinned 1.93 available:

- `cargo test -p formualizer-eval` — 2393 passed, 1 failed. That failure is `run_formula_test_suite` at `914 passed / 16 skipped / 3 failed`, identical to unmodified `main`; the three cases are complex-number formatting, unrelated to this change.
- `cargo fmt --all` — clean.
- `cargo clippy -p formualizer-eval --all-targets` — no warnings in the touched file. (The crate doesn't pass `-D warnings` under clippy 1.96 on `main` either — 22 errors before this change — which I take to be drift from the pinned toolchain.)
- Verified end-to-end through the WASM bindings on a real workbook.

## Note on scope

I originally held this back because #199 is reworking `engine/eval.rs` and `arrow_store`, and this touches shared plumbing. Sending it since it's built and verified, but entirely happy for you to redirect the approach or fold it into that work.

Related: #218 fixes `FILTER`, which fails for a separate reason. Both are needed for `SUM(FILTER(C2:C5,B2:B5="Jakarta"))` — that one lets `FILTER` produce its array, this one lets `SUM` consume it.


maulafaizz:
Force-pushed an extension, flagging it in case you were mid-review.

The original version only handled `CalcValue::Range`. Testing through the WASM bindings turned up a second form: a range comparison or elementwise operation (`B2:B5="Jakarta"`, `(B2:B5="x")*1`) evaluates to `CalcValue::Scalar(LiteralValue::Array)` instead, so those arrays still failed to reach any aggregate — `SUM((B2:B5="Jakarta")*1)` returned `#VALUE!`.

Same fallback, one more match arm, plus a second test covering that form. I folded it in here rather than opening another PR since it is the same function and the same bug; without it this fix is only half of #211, which would have been a confusing state to leave the issue in.

Full suite still matches `main` at `914/16/3`.

PSU3D0:
Confirmed in both FormulaPlane Off and AuthoritativeExperimental modes: this fixes the reported computed-array aggregate cases. Before merge, we plan to make the shared path select reference versus computed-value evaluation structurally, preserve exact `OFFSET`/`INDIRECT` errors, avoid rematerializing an existing `RangeView`, and add AST/arena plus volatile-evaluation coverage.

The broader coordination note is on #213. If you would like to take another pass at that path selection before our maintainer work later this week, leave a note here; otherwise, no action is required. This should land before #218.

PR #218, opened 2026-07-21
fix(eval): stop FILTER rejecting a computed include argument

### LINKED ISSUE #216: [bug] FILTER returns #REF! in all forms, including with a literal boolean array

## Summary

`FILTER` returns `#REF!` in every form I tried, including standalone and with a literal boolean array as the `include` argument. `SORT`, `UNIQUE`, `SEQUENCE` and `TRANSPOSE` on the same fixture all behave correctly, so this looks specific to `FILTER`.

## Reproduction

```js
await initializeWasm();

const T = (v) => ({ type: 'Text', value: v });
const N = (v) => ({ type: 'Number', value: v });
const cells = [];
[['Ani','Jakarta',10],['Budi','Bandung',20],['Cici','Jakarta',30],['Dodi','Surabaya',40]]
  .forEach((r, i) => cells.push(
    { row: i + 2, col: 1, value: T(r[0]) },
    { row: i + 2, col: 2, value: T(r[1]) },
    { row: i + 2, col: 3, value: N(r[2]) },
  ));

const wb = Workbook.fromJson(JSON.stringify({ sheets: { S: { cells } } }));
const F = [
  '=FILTER(C2:C5,B2:B5="Jakarta")',
  '=INDEX(FILTER(C2:C5,B2:B5="Jakarta"),1)',
  '=SUM(FILTER(C2:C5,{TRUE;FALSE;TRUE;FALSE}))',
  '=ROWS(B2:B5="Jakarta")',
  '=SUMPRODUCT((B2:B5="Jakarta")*C2:C5)',
  '=SUM(SORT(C2:C5,1,-1))',
];
F.forEach((f, i) => wb.setFormula('S', 20 + i * 3, 8, f));
wb.evaluateAll();
F.forEach((f, i) => console.log(f.padEnd(46), '->', JSON.stringify(wb.evaluateCell('S', 20 + i * 3, 8))));
```

## Actual

```
=FILTER(C2:C5,B2:B5="Jakarta")                 -> "#REF!"     Excel: {10;30}
=INDEX(FILTER(C2:C5,B2:B5="Jakarta"),1)        -> "#REF!"     Excel: 10
=SUM(FILTER(C2:C5,{TRUE;FALSE;TRUE;FALSE}))    -> "#REF!"     Excel: 40
=ROWS(B2:B5="Jakarta")                         -> 4           correct
=SUMPRODUCT((B2:B5="Jakarta")*C2:C5)           -> 40          correct
=SUM(SORT(C2:C5,1,-1))                         -> 100         correct
```

## What this rules out

- **Not comparison broadcasting.** `ROWS(B2:B5="Jakarta")` returns 4 and `SUMPRODUCT((B2:B5="Jakarta")*C2:C5)` returns 40, so the range comparison does produce a usable 4-row boolean array.
- **Not the `include` argument being a computed expression.** Passing a literal `{TRUE;FALSE;TRUE;FALSE}` array fails identically.
- **Not dynamic-array support generally.** `SORT`, `UNIQUE`, `SEQUENCE` and `TRANSPOSE` all spill and aggregate correctly on the same workbook.

`FilterFn::eval` in `crates/formualizer-eval/src/builtins/lookup/dynamic.rs` takes `args[0].range_view()?` and `args[1].range_view()?`, and its `arg_schema` declares both as `ShapeKind::Range` with `by_ref: true` — that pairing may be worth a look, though I haven't traced which of the two calls produces the `#REF!`.

Related to but distinct from #211; I originally mis-filed this as an instance of that bug and have corrected the record there.

maulafaizz:
Fixes #216.

## Problem

`FILTER` declares its `include` argument as `by_ref: true`. Argument preparation rejects a by-ref argument that doesn't resolve to a bare reference (`args.rs`, *"By-ref argument: require a reference"*), returning `#REF!` **before** `FilterFn::eval` runs.

That made the working case the unusual one:

```
=FILTER(C2:C5,E2:E5)                         -> worked   (bare reference)
=FILTER(C2:C5,B2:B5="Jakarta")               -> #REF!    Excel: {10;30}
=SUM(FILTER(C2:C5,{TRUE;FALSE;TRUE;FALSE}))  -> #REF!    Excel: 40
```

In Excel the `include` argument is nearly always a computed boolean array rather than a reference, so requiring a reference is the wrong constraint. The existing `filter_basic_and_if_empty` test only ever passes a reference, which is why this went unnoticed.

## Change

Drop `by_ref` on that argument so it resolves through the value path, which `range_view()` already materialises. `shape` stays `ShapeKind::Range`.

Updated the `[formualizer-docgen:schema]` block to match, since `arg_schema` changed. I edited it by hand: `cargo run -p xtask -- schema --apply` reports `files scanned: 0` / `entries touched: 0` for me and rewrites nothing, so it could not do it. Worth a look separately — it means the generator is currently a no-op rather than that this block is exempt.

## Tests

Added `filter_accepts_computed_include`, passing an inline `{TRUE;FALSE;TRUE}` array rather than a reference. I confirmed it fails with `by_ref: true` restored and passes with the fix, so it pins the actual behaviour.

## Verification

Rust 1.96.1 (nixpkgs):

- `cargo test -p formualizer-eval` — 2392 passed, 1 failed; the failure is `run_formula_test_suite` (`914/16/3`), identical to unmodified `main` and unrelated (complex-number formatting).
- `cargo fmt --all` — clean.
- `cargo clippy -p formualizer-eval --all-targets` — no warnings in the touched file.
- `xtask -- schema` — no stale entries; `xtask -- audit --strict` — no issues. Both report scanning nothing (`files scanned: 0`, `total registered: 0`), so treat that line as a no-op rather than as coverage.
- Verified end-to-end through the WASM bindings.

## Note

Independent of #217 (issue #211) — different file, different cause. Both are needed for `SUM(FILTER(...))` to work: this one lets `FILTER` produce its array, that one lets `SUM` consume it.


PSU3D0:
Confirmed: removing the incorrect `by_ref` constraint unlocks inline Boolean-array masks. The primary computed-comparison case also needs the safe computed-value path from #217, so our plan is to harden #217 first, then rebase this PR and add an actual binary-mask test, AST/workbook coverage, `SUM(FILTER(...))` integration, and refreshed generated metadata.

The broader coordination note is on #213. If you would like to take another pass after #217’s direction is settled, leave a note here; otherwise, no action is required before our maintainer pass later this week.

PSU3D0:
Thanks for this — your diagnosis of the argument-preparation rejection was exactly right, and it led somewhere bigger than the original report.

While verifying the fix I found that flipping `include` off `by_ref` was necessary but not sufficient. `ArgumentHandle::range_view` also required a reference for `Function`/`BinaryOp` nodes, so the idiomatic form in the PR title still failed:

`=FILTER(A1:A3,B1:B3="x")` → `#REF!`

Your test passed because an inline `{TRUE;FALSE;TRUE}` array literal takes a different branch in `range_view` than a computed comparison does.

The same defect affected sixteen data arguments across the whole dynamic-array family, not just FILTER. Confirmed broken before this change:

`=TRANSPOSE({1,2})`, `=SORT({3;1;2})`, `=UNIQUE({1;1;2})`, `=SORT(SEQUENCE(3))`, `=TAKE({1;2;3},2)`, `=DROP({1;2;3},1)`, `=SORTBY(...)`, `=XLOOKUP(20,A1:A3*1,A1:A3)`, `=XMATCH(20,A1:A3*1)`, plus GROUPBY and PIVOTBY.

All are valid in Excel and all returned `#REF!`.

So I extended the change on your branch:

- Removed `by_ref` from all sixteen data arguments. `by_ref` now means "this argument has genuine reference semantics" — `OFFSET`, `ROW` and `COLUMN` keep it.
- Rewrote `range_view` to delegate to the cached `resolve_once` path introduced in #217, so references keep their lazy views, computed arguments resolve through one evaluation, and genuine reference failures stay errors instead of being masked.

The underlying model is the one established in #217: a reference is a capability, not a requirement. `by_ref` was only ever acting as a rejection gate, since `dispatch` discards the prepared arguments and `eval` re-resolves anyway.

Added coverage: all sixteen functions with computed inputs under both FormulaPlane modes, whole-column references, preserved `OFFSET`/`INDIRECT` errors, and a counting function asserting a computed argument is evaluated exactly once.

One thing deliberately left alone: Excel treats `=TRANSPOSE(2)` as a 1x1 array, while we still reject computed scalars. That touches every range-consuming builtin, so it belongs in its own change; there is a test documenting the current behavior rather than endorsing it.

Your commit is preserved and this lands with your authorship. Closes #216 as well.

Issue #240, opened 2026-07-27
FormulaPlane authoritative mode is ~cubic on bulk XLSX load: 963x slower first eval at 26k cells


PSU3D0:
`FormulaPlaneMode::AuthoritativeExperimental` is roughly cubic in the number of formula cells when a workbook is **loaded** (`LoadStrategy::EagerAll`), while being free when the identical formulas are added incrementally via `set_formula`. At 26k formula cells the first evaluation takes **8.6 minutes** against 0.5 seconds with the plane off.

Values are unaffected — 0 mismatches across 26,000 cells in every run — so this is purely a performance defect, but it is large enough to block any authoritative cutover.

## Measurement

`probe-fp-coverage` (standing coverage probe, mixed 13-section corpus), `main` at 9b8fb52a, release build:

| formula cells | `Off` first eval | `AuthoritativeExperimental` first eval | ratio |
|---|---|---|---|
| 1,625 | 9 ms | 26 ms | 2.9x |
| 3,250 | 20 ms | 155 ms | 7.8x |
| 6,500 | 56 ms | 7,420 ms | 132x |
| 13,000 | 161 ms | 66,658 ms | 414x |
| 26,000 | 537 ms | 516,976 ms | 963x |

Doubling the corpus multiplies authoritative first-eval by roughly 8x, i.e. `O(n^3)`. `Off` stays near-linear over the same range.

Load time itself is *not* affected (542 ms authoritative against 565 ms off at 26k cells), and warm recalc is *not* affected (3 ms against 2 ms). The cost is isolated to the first evaluation after a bulk load.

## Localisation

The regression is in the ingest route, not in span evaluation. Driving the **same generated corpus** (`formualizer_testkit::fp_coverage::generate`) into a workbook cell-by-cell with `set_formula` instead of loading it from XLSX removes the problem completely:

| rows/section | in-memory `set_formula`, authoritative | XLSX `EagerAll`, authoritative |
|---|---|---|
| 250 | 38 ms (1.0x vs off) | 155 ms (7.8x) |
| 500 | 140 ms (1.1x) | 7,420 ms (132x) |
| 1,000 | 451 ms (1.0x) | 66,658 ms (414x) |
| 2,000 | 1,711 ms (1.0x) | 516,976 ms (963x) |

Same formulas, same sheets, same named ranges, same mode. 1.7 seconds against 517 seconds.

[...]

PR #249, opened 2026-07-30
perf(eval): retain bounded topology prefixes

### LINKED ISSUE #244: Topology cache overflow discards work and rebuilds per request: revisit the O(n^2) compile and its 100k cap

PSU3D0:
## Summary

- retain a bounded mixed-topology prefix when the candidate or edge limit is reached instead of discarding all successful compile work
- record exactly which source producers are complete and combine those cached relationships with exact paged discovery for the uncovered tail on later requests
- keep retained-byte accounting authoritative, including the complete-source hash capacity, mirrored precedent adjacency, producer/read indexes, bindings, fallback capacity, and all vector capacities
- make the coverage probe perform two real value-edit recalculations and report cache outcomes, strategies, cap hits, and cache-hit telemetry; its previous “warm” call was a no-dirty no-op and could not demonstrate this issue

## Design decision

I chose bounded partial retention plus an exact tail.

The compiler still stops at the existing candidate, edge, or retained-byte bound. A source is added to `complete_sources` only after every relationship for that source was compiled. Candidate overflow publishes no partial query result, and an edge-overflow source is not marked complete. On a retained cache hit, paged scheduling uses cached relationships only for complete sources and runs the indexed exact query for every other source. The schedule therefore has the same authority as the exact path; the partial prefix is never treated as complete demand authority.

The first request continues to use the exact request path after compilation. Reuse starts on a later cache hit, avoiding additional work on the cold path and keeping partial-cache correctness isolated from initial publication. If an initial request cannot retain a prefix, a later stable generation may build it; this is visible at N=1000 and N=2000 below as `skipped_overflow -> built -> hit`.

A partial result is retained only when it contains useful relationships and its full retained estimate fits. Zero limits, allocation failure, unsupported projection, missing result regions, and retained-byte overflow still produce a typed atomic cache skip. Dynamic legacy dependencies still bypass retention. No admission, authority, or FormulaPlane placement semantics changed.

## Alternatives rejected

- **Raise or remove 100,000:** rejected. The growing-range relationships are genuine, and removing the bound violates resource accounting. Raising an uncalibrated constant merely moves the cliff.
- **Replace the three bounds with one cost score:** deferred. Candidate count remains a direct compile-work guard, edge count bounds relationship work, and retained bytes bound ownership. A combined weighting needs C6 calibration across workbook shapes; inventing weights here would be less defensible than retaining all three independent limits.
- **Cache an overflow marker only:** rejected because it would avoid repeated compilation but still rebuild every source relationship per request. The complete-source prefix lets paged discovery rebuild only the tail.
- **Use a partial topology as demand authority:** rejected. Backward demand closure needs complete precedent coverage. Target demand continues through the exact strategies unless the topology is complete.
- **Read-table deduplication:** already disproven by #243/#248. It reduces bookkeeping but not the genuine quadratic source-to-consumer relationships and regressed first evaluation.

## Provenance of `100_000`

`git log -S'max_formula_plane_cache_candidates'` traces the value to `3cca32e8` / merged commit `b33227bd` (`perf(eval): cache mixed FormulaPlane topology`, PR #186). The implementation and PR described it as a hard bound, but neither contains calibration data or a rationale for 100,000 specifically. I therefore treat it as an uncalibrated default and leave it unchanged.

## Measurements

Release binary SHA-256 after the final rebuild: `898b20edc21f1f906a564137ee9bd237953b4fd93ece4361c0844346d59315dc`.

The probe now performs: cold evaluation, edit `+1` and warm evaluation, edit `+2` and warm-hit evaluation. Edits are outside the timed evaluation windows. Every row below is one process/sample in milliseconds.

### Branch (`cc783ec2`)

| N | sample | Off first | Auth first | Off warm 1 | Auth warm 1 | Off warm hit | Auth warm hit | first / warm 1 / warm-hit cache outcome |
|---:|---:|---:|---:|---:|---:|---:|---:|---|
| 500 | 1 | 56 | 513 | 0 | 1 | 1 | 1 | built / hit / hit |
| 500 | 2 | 57 | 520 | 1 | 1 | 0 | 2 | built / hit / hit |
| 500 | 3 | 56 | 523 | 0 | 1 | 0 | 0 | built / hit / hit |
| 1000 | 1 | 162 | 690 | 1 | 756 | 1 | 1 | skipped_overflow / built / hit |
| 1000 | 2 | 164 | 689 | 1 | 706 | 1 | 1 | skipped_overflow / built / hit |
| 1000 | 3 | 158 | 678 | 1 | 733 | 1 | 1 | skipped_overflow / built / hit |
| 2000 | 1 | 541 | 3030 | 2 | 1796 | 2 | 2 | skipped_overflow / built / hit |
| 2000 | 2 | 544 | 3040 | 2 | 1767 | 2 | 2 | skipped_overflow / built / hit |
| 2000 | 3 | 540 | 3030 | 2 | 1760 | 2 | 2 | skipped_overflow / built / hit |

All branch warm-hit requests reported `exact_paged_indexed`, `cache_hit_events=1`, and exact values. N=500 reported a candidate cap hit and retained the prefix on the cold request. At N=1000 and N=2000 the cold request could not retain that generation; the first edited request built the retained partial generation and the next request hit it.

### Contemporary main (`68cd42c4`), same real first edit

| N | sample | Off first | Auth first | Off warm 1 | Auth warm 1 | first / warm cache outcome |
|---:|---:|---:|---:|---:|---:|---|
| 500 | 1 | 54 | 501 | 0 | 368 | skipped_overflow / skipped_overflow |
| 500 | 2 | 54 | 528 | 0 | 368 | skipped_overflow / skipped_overflow |
| 500 | 3 | 55 | 494 | 0 | 379 | skipped_overflow / skipped_overflow |
| 1000 | 1 | 165 | 682 | 1 | 713 | skipped_overflow / skipped_overflow |
| 1000 | 2 | 157 | 681 | 1 | 706 | skipped_overflow / skipped_overflow |
| 1000 | 3 | 161 | 677 | 1 | 705 | skipped_overflow / skipped_overflow |
| 2000 | 1 | 544 | 3009 | 2 | 1756 | skipped_overflow / skipped_overflow |
| 2000 | 2 | 574 | 3052 | 2 | 1782 | skipped_overflow / skipped_overflow |
| 2000 | 3 | 557 | 3006 | 2 | 1753 | skipped_overflow / skipped_overflow |

Cold medians are branch/main 520/501, 689/681, and 3030/3009 ms. These small 19 ms (3.8%), 8 ms (1.2%), and 21 ms (0.7%) deltas sit inside the observed cold run spread and machine drift; N=500 remains below the previously verified 527 ms baseline. The material result is the removal of the perpetual warm cliff: main continues to report `skipped_overflow`, while the branch reaches a cache hit and evaluates the second edit in 0–2 ms at N=500, 1 ms at N=1000, and 2 ms at N=2000. I did not trade a substantial cold regression for this result; the exact cold path is retained, and contemporary first-eval distributions overlap.

Coverage was 69.2% with 9 spans in all 9 required branch probe runs, and `value_equality.mismatches` was 0 in every run after both edits.

The corpus changes from complete cached topology at N=315 to bounded partial topology at N=320. N=300, 310, and 315 reported `compiled_and_cached`; N=320, 325, 330, 350, and 500 reported `exact_paged_indexed` with a candidate cap hit while retaining the prefix.

## Fault matrix

The existing boundary coverage was extended rather than replaced:

- candidate limit 0: still `CacheSkipped(MaxCandidatesExceeded)`, no cache installed
- two-candidate fixture with limit 1 (cap+1 observed): retains one complete source, marks the topology incomplete, and the hybrid schedule is layer-identical to the full exact schedule while issuing exact queries only for uncovered sources
- same fixture with limit 2 (at cap): complete topology is cached
- edge limit 0: still a typed cache skip
- retained-byte limit 0: still a typed cache skip with zero retained relationships
- perpetual zero-limit skip: existing three-request strategy matrix still preserves the terminal value, increments the request streak 1/2/3 and aggregate current/max streak to 3, and leaves scratch at zero after every request

The zero-limit atomic assertion was not relaxed. The new test changes the expected behavior only when there is a useful, fully covered prefix that fits every retained bound.

## Verification

- `cargo fmt --all --check`
- `cargo test -p formualizer-eval --lib --all-features` — 2447 passed, 0 failed, 12 ignored (one new boundary test over the 2446 baseline)
- 30-run `formula_plane_ingest_shadow --test-threads=32` soak — 0 failures

Closes #244

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,7 @@ All notable changes to Formualizer will be documented in this file.
 
 ### Performance
 
+- Mixed FormulaPlane topology overflow now retains the bounded, accounted prefix and its indexes instead of discarding successful compile work. Paged schedule discovery reuses complete cached sources and derives only the uncovered tail exactly, so dense growing-range workbooks avoid rebuilding the same overflowed topology on every recalculation.
 - Experimental authoritative FormulaPlane evaluation now uses the existing consumer-read interval index when exact schedule construction falls back after topology-cache overflow, avoiding full-table scans per dirty formula during the first evaluation of bulk-loaded workbooks. (#240)
 - Experimental authoritative FormulaPlane evaluation now caches accounted consumer and precedent topology across warm and value-only evaluations. Exact graph, authority, semantic, provider, and dynamic-reference revisions invalidate or bypass stale cache generations; candidate, edge, and retained-byte cache overflow selects exact paged/run/native/repeated-pass request topology without span demotion, while span-free and warm no-dirty requests retain topology-free sparse paths.
 - Proven complete source-formula families now parse and analyze one anchor and avoid per-descendant strings, ASTs, staging entries, and graph vertices. In same-machine release probes, a clean 100k-family load improved from 997 ms and 313 MiB RSS under forced replay to 129 ms and 26 MiB RSS; a 1M-family load completed in 2.2 s at 167 MiB RSS instead of 13.1 s at 3.0 GiB.
--- a/crates/formualizer-bench-core/src/bin/probe-fp-coverage.rs
+++ b/crates/formualizer-bench-core/src/bin/probe-fp-coverage.rs
@@ -71,6 +71,16 @@ mod probe {
         load_ms: u128,
         eval_first_ms: u128,
         eval_warm_ms: u128,
+        eval_warm_hit_ms: u128,
+        topology_cache_outcome_first: String,
+        topology_strategy_first: String,
+        topology_candidate_cap_hits_first: u64,
+        topology_cache_outcome_warm: String,
+        topology_strategy_warm: String,
+        topology_cache_hit_events_warm: u64,
+        topology_cache_outcome_warm_hit: String,
+        topology_strategy_warm_hit: String,
+        topology_cache_hit_events_warm_hit: u64,
     }
 
     #[derive(Debug, Serialize)]
@@ -193,10 +203,71 @@ mod probe {
         let t = Instant::now();
         wb.evaluate_all()?;
         let eval_first_ms = t.elapsed().as_millis();
+        let (
+            topology_cache_outcome_first,
+            topology_strategy_first,
+            topology_candidate_cap_hits_first,
+        ) = wb
+            .engine()
+            .last_evaluation_resource_request_stats()
+            .map(|stats| {
+                (
+                    stats.topology.cache_outcome.as_str().to_string(),
+                    stats.topology.strategy.as_str().to_string(),
+                    stats.topology.candidate_cap_hits,
+                )
+            })
+            .unwrap_or_else(|| ("not_used".to_string(), "not_used".to_string(), 0));
 
+        if let Some(input) = corpus.data_values.first() {
+            wb.set_value(
+                input.sheet,
+                input.row,
+                input.col,
+                LiteralValue::Number(input.value + 1.0),
+            )?;
+        }
         let t = Instant::now();
         wb.evaluate_all()?;
         let eval_warm_ms = t.elapsed().as_millis();
+        let (topology_cache_outcome_warm, topology_strategy_warm, topology_cache_hit_events_warm) =
+            wb.engine()
+                .last_evaluation_resource_request_stats()
+                .map(|stats| {
+                    (
+                        stats.topology.cache_outcome.as_str().to_string(),
+                        stats.topology.strategy.as_str().to_string(),
+                        stats.topology.cache_hit_events,
+                    )
+                })
+                .unwrap_or_else(|| ("not_used".to_string(), "not_used".to_string(), 0));
+
+        if let Some(input) = corpus.data_values.first() {
+            wb.set_value(
+                input.sheet,
+                input.row,
+                input.col,
+                LiteralValue::Number(input.value + 2.0),
+            )?;
+        }
+        let t = Instant::now();
+        wb.evaluate_all()?;
+        let eval_warm_hit_ms = t.elapsed().as_millis();
+        let (
+            topology_cache_outcome_warm_hit,
+            topology_strategy_warm_hit,
+            topology_cache_hit_events_warm_hit,
+        ) = wb
+            .engine()
+            .last_evaluation_resource_request_stats()
+            .map(|stats| {
+                (
+                    stats.topology.cache_outcome.as_str().to_string(),
+                    stats.topology.strategy.as_str().to_string(),
+                    stats.topology.cache_hit_events,
+                )
+            })
+            .unwrap_or_else(|| ("not_used".to_string(), "not_used".to_string(), 0));
 
         let mut values = Vec::new();
         for cell in corpus.sections.iter().flat_map(|s| s.formulas.iter()) {
@@ -233,6 +304,16 @@ mod probe {
                 load_ms,
                 eval_first_ms,
                 eval_warm_ms,
+                eval_warm_hit_ms,
+                topology_cache_outcome_first,
+                topology_strategy_first,
+                topology_candidate_cap_hits_first,
+                topology_cache_outcome_warm,
+                topology_strategy_warm,
+                topology_cache_hit_events_warm,
+                topology_cache_outcome_warm_hit,
+                topology_strategy_warm_hit,
+                topology_cache_hit_events_warm_hit,
             },
             values,
             coverage,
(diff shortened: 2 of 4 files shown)


PSU3D0:
Reviewed independently and merging. Two notes from verification, one of which is useful for a follow-up.

## The win holds under an apples-to-apples control

Because this PR also changes the measuring instrument, I built a control — **`main`'s engine with this PR's probe** — so both sides are measured the same way:

| N | `main` + new probe, `warm_hit` | this branch, `warm_hit` | outcome on `main` | outcome here |
|---|---|---|---|---|
| 500 | 351 ms | **0 ms** | `skipped_overflow` | `hit` |
| 1000 | 701 ms | **1 ms** | `skipped_overflow` | `hit` |
| 2000 | 1,818 ms | **3 ms** | `skipped_overflow` | `hit` |

Cold first-eval is within noise (499→518, 670→680, 3016→3003; +3.8%, +1.5%, −0.4%), coverage stays 69.2% with 9 spans, and mismatches are 0 everywhere. Edited recalculation on a 26k-cell workbook going from ~1.8 s to 3 ms is the interactive-editing case, so this is the most user-visible result of the whole effort.

## Fixing the probe was the right call and deserves emphasis

The pre-existing `warm` measurement was a no-dirty no-op, so every "warm" number reported across this effort — including the `2 ms` vs `3 ms` figures in #240 — was measuring nothing at all. Adding a real edit plus cache telemetry corrects the record, and it is what made this PR's win visible in the first place. Good catch.

## One observation for a follow-up

I mutation-tested the retention path two ways:

- Forcing `reuse_partial_topology = false`: **all 2447 tests still passed** and the measured win was unchanged.
- Making `schedule_dirty_work_paged_hybrid` ignore the retained prefix entirely: `candidate_cap_retains_a_safe_prefix_and_exactly_schedules_the_tail` **correctly failed** — so the prefix-consumption logic is genuinely test-protected. But `warm_hit` at N=2000 stayed at **4 ms** against `main`'s 1,818 ms.

So the timing win comes primarily from **no longer recompiling the topology on every request**, not from consuming the prefix contents. That is a coherent and valuable result — retaining any accounted compile output flips the next request to `Hit` and skips the recompile — and correctness is unaffected. It does mean the headline behaviour is not directly pinned by a test: a change that stopped retaining on overflow would keep CI green while silently restoring the per-request recompile.

Worth a small engine-level test asserting that a second request after candidate overflow reports `Hit` rather than `skipped_overflow`. Filing that as a follow-up rather than holding the merge, since the behaviour is verified and the correctness invariants are covered.

Verified locally: `fmt` clean, `clippy -D warnings` clean, wasm clippy clean, 2447 passed / 0 failed, module soak 0/30, and CI green on all five jobs.

Issue #171, opened 2026-07-06
Span delete-compaction rule-frame: verify relative offsets stay authored−origin through head/tail-trimmed compactions


PSU3D0:
Follow-up flagged during the #169 frame-invariant audit (see the "frame invariant" documentation at `rewrite_frame_is_sound` in `crates/formualizer-eval/src/formula_plane/structural_shift.rs`).

Span read-summary projection rules store placement-relative offsets that must equal `authored_coordinate − template_origin` at all times. #169 fixed two violations of this (origin-follows shifts now transform rule offsets via `DirtyProjectionRule::with_relative_offsets_shifted`; the rewrite arm re-derives its summary wholesale). The delete-compaction path was intentionally NOT touched: `compact_read_summary_through_delete` + `origin_through_delete` keep rules untransformed while a delete inside or at the edge of a span's read region mixes shifted and unshifted placements — the rule-frame story there is not a pure translation and needs its own investigation.

Current exposure is bounded by two guards from #165/#169: compaction demotes when the delete displaces an absolute read (`summary_has_displaced_absolute_read`) and when bindings are not single-literal. The open question is the relative-offset flavor for head-trimmed compactions: whether any reachable delete shape leaves a compacted span whose stored offsets diverge from `authored − origin`, which would silently mis-route incremental dirtiness (stale values, no error).

Suggested approach: extend the W0 oracle harness (`formula_plane_structural_w0_oracle.rs`, the "structural op → incremental write → evaluate" probes) with delete shapes that trim the head/tail of the read region while the span survives compaction; any red is a confirmed bug, and the fix likely mirrors `with_relative_offsets_shifted` with per-placement care.

PR #252, opened 2026-07-30
perf(eval): contract isolated legacy formula islands

### LINKED ISSUE #251: One span anywhere routes the whole workbook through the mixed planner: 7.7x on mixed workbooks
Authoritative mode currently pays large overhead on regions it cannot accelerate, because the legacy fast-path early-out is **global** rather than per-sheet.

## Measurement

`probe-fp-coverage --rows-per-section 1000`, on `main` @ `699a129a` (i.e. after #247 and #249):

| sections | coverage | spans | `Off` | `Authoritative` | ratio |
|---|---|---|---|---|---|
| `anchored_agg` alone | 100% | 1 | 1 ms | **0 ms** | 0.0x — faster |
| `running_total` alone | 100% | 1 | 4 ms | **1 ms** | 0.2x — faster |
| `self_cumulative` alone | 0% | 0 | 72 ms | **73 ms** | 1.0x — no overhead |
| **all three together** | 66.7% | 2 | 80 ms | **619 ms** | **7.7x slower** |

Each subset in isolation is fine or better. Only the *mixture* is pathological.

## Why

`Engine::evaluate_authoritative_formula_plane` has a global early-out:

```rust
// engine/eval.rs
if self.graph.formula_authority().active_span_count() == 0 {
    // pure legacy evaluator: compressed-range scheduling, no mixed topology
```

`self_cumulative` (`=SUM($C$1:$C{r-1})*0+B{r}`) is rejected with `InternalDependency`, so alone it admits zero spans, hits this early-out, and runs at legacy speed. Add *one* span anywhere in the workbook and the same 1,000 legacy cells are routed through the mixed-topology planner instead — where a growing-range read pattern generates about n²/2 relationships.

So the cost is not a property of the formulas, and not a complexity problem. It is a routing decision made at workbook granularity for a property that is regional.

## Why this is the highest-value remaining perf item

The exponent is not the problem. Measured scaling per doubling on the same corpus:

- Linear-only sections: `Off` 2.27x, `Authoritative` 2.60x — **linear, as expected**.
- Growing-range sections: `Off` 3.80x, `Authoritative` 4.59x — **quadratic in both modes**, because `SUM($A$1:$A{r})` for r = 1..n is Σr ≈ n²/2 cell reads. That work is inherent to the formulas; Excel pays it too.

So the engine has no systemic quadratic of its own. What remains is a ~7.7x constant factor on mixed workbooks, and it comes from this routing choice. Real workbooks are mixed almost by definition — a running-total column beside copied-formula columns is an ordinary spreadsheet — so this is the common case, not a corner.

## Suggested approach

Make the early-out granular instead of global: evaluate a sheet (or a region) through the legacy path when it has no admitted spans, even while other sheets do. The decision data already exists — span authority is tracked per sheet and `Region` carries `sheet_id` throughout.

Constraints that must hold:
- Values, errors and ordering must be identical; this is a scheduling choice, not a semantic one.
- Cross-sheet dependencies between a span-bearing sheet and a legacy-only sheet must still be discovered. This is the part to design carefully, and the reason it is not a one-line change.
- Coverage must not move (69.2% / 9 spans on the corpus) — this changes routing, not admission.
- Resource accounting must still charge whichever path runs.

Acceptance signal: the three-section mixture above should approach the sum of its parts (roughly 73 ms, not 619 ms), with `value_equality.mismatches` at 0 and coverage unchanged.

PSU3D0:
## Summary

- contract proven span-disconnected legacy work into a non-finalising legacy scheduler primitive under the existing FormulaPlane coordinator
- retain span-connected legacy producers in the mixed topology and preserve mixed-schedule cycle detection/demotion
- add bounded route events and accounted island/topology shape telemetry
- fail closed for dynamic references, names, spills/array formulas, unsupported read summaries, boundary-discovery overflow, and structural axis edits while #171 remains open

## Isolation proof

The retained mixed topology starts with every active span result and read summary. Boundary discovery finds both directions: span result to legacy consumer and legacy result to span consumer. It then computes an undirected fixed point through legacy result/read relationships. Every legacy producer in that closure remains in the existing mixed scheduler. Therefore any path from a legacy producer to a virtual span member, including a path that closes a cycle through the span, remains visible to the producer-bounded mixed cycle detector and follows the unchanged G8 demotion path before legacy Tarjan runs.

Only the complement is contracted. It is evaluated through `create_evaluation_schedule` and the existing layer/SCC walkers, but the primitive does not begin a request, lease or acknowledge dirty authority, redirty volatiles/iterative SCCs, refresh the clock, or advance `recalc_epoch`. Those operations remain owned once by the outer FormulaPlane coordinator.

A stricter fast proof handles the measured multi-sheet corpus without constructing irrelevant legacy indexes: every legacy result and every legacy read sheet must be disjoint from the union of all span result and span read sheets. Same-sheet work never uses that shortcut and goes through exact regional boundary discovery.

## Fail-closed conditions

- any dynamic legacy formula
- any workbook- or sheet-scoped name
- any active spill or legacy array formula
- a non-cell dependency or a range that cannot be represented by a structural read region
- a boundary query, closure query, allocation, candidate cap, edge cap, or retained-byte cap that cannot complete exactly
- any row/column structural edit after engine creation, because #171 leaves relocated summaries unsuitable as a disconnection proof
- no non-empty isolated membership

On all of these paths the existing global mixed coordinator and its exact overflow ladder remain in use.

## Request lifecycle and observability

`evaluate_authoritative_formula_plane` no longer begins a request a second time. Tests pin one begin/clock refresh and one epoch increment per public request while a legacy iterative accumulator runs repeatedly beside an unrelated active span.

Route telemetry is fixed at 16 inline events and records planned/executed phase, deterministic island ID, a bounded sorted sheet set, route, authority/index epochs, transition reason, and demotion/replan generations. It also reports dropped events, inline event bytes, retained membership bytes, omitted legacy relationships, and retained boundary relationships. Membership bytes are included in topology retained-memory accounting and the cache key explicitly carries island and boundary revisions.

## Correctness and regression tests

- hardcoded cross-sheet post-warm back-edge values/errors prove mixed detection demotes the span before legacy Tarjan
- repeated hardcoded iterative accumulator values prove persistence and exactly one begin/clock sample plus one epoch increment per request
- same-sheet disconnected work asserts an executed contraction route; the connected variant asserts no contraction
- the existing mixed-tail fault, cancellation, cache cap, retained plan, dirty lease, demotion, target, delta, and structural suites remain green
- no existing assertion was relaxed; the established `producers_observed`/candidate/edge telemetry meaning is preserved while omissions are exposed separately

## Mutations

- disabling both contraction proofs fails `mixed_tail_reads_complete_in_one_authoritative_pass` at the named executed-route assertion
- clearing the span-connected closure before classification fails `cached_mixed_topology_matches_off_first_warm_and_post_edit` at the named connected-island assertion

Both mutations exited 101 with exactly the intended named test failure.

## Measurements

Release binary hash changed from `02abd536d9fc4784b31dca87021611ed5ec99516fe75befb5e1077cd221039ec` to `a10f5f2283d8e98f133f8c00d6a8c82d8ff3e0dad1e7d42f46c0838d45fe2e28` before final measurements.

Three-section mixture at N=1000, five cold-process samples:

| run | Off first | authoritative first | warm hit | mismatches |
|---|---:|---:|---:|---:|
| 1 | 78 ms | 73 ms | 0 ms | 0 |
| 2 | 78 ms | 73 ms | 0 ms | 0 |
| 3 | 79 ms | 73 ms | 0 ms | 0 |
| 4 | 79 ms | 77 ms | 0 ms | 0 |
| 5 | 84 ms | 79 ms | 0 ms | 0 |

This moves the reported 619 ms authoritative result to a 73 ms median.

Full corpus:

| N | coverage | spans | mismatches | authoritative first | authoritative warm hit |
|---:|---:|---:|---:|---:|---:|
| 500 | 69.2% | 9 | 0 | 520 ms median (3 paired samples; main 531 ms) | 0 ms |
| 1000 | 69.2% | 9 | 0 | 689 ms (paired main 699 ms) | 1 ms |
| 2000 | 69.2% | 9 | 0 | 3104 ms median (10 paired samples; main 3096 ms) | 3 ms |

At N=2000 the paired medians differ by 0.2% with alternating wins and are within run noise; warm work did not regress. The supplied historical first-eval baseline was 3059 ms.

## Validation

- `cargo fmt --all --check`
- `cargo test -p formualizer-eval --lib --all-features`: 2449 passed, 0 failed, 12 ignored
- `formula_plane_ingest_shadow --test-threads=32`: 30/30 module runs passed
- `formula_plane_mixed_legacy_tail_reads`: 16 passed, including the demote-loop regression net

Closes #251

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,7 @@ All notable changes to Formualizer will be documented in this file.
 
 ### Performance
 
+- Experimental authoritative FormulaPlane evaluation now contracts proven span-disconnected legacy islands into the compressed legacy scheduler while retaining span-connected legacy producers in the mixed cycle detector. Dynamic references, names, spills, structural-summary uncertainty, and boundary-discovery overflow fail closed to the global mixed planner. (#251)
 - Mixed FormulaPlane topology overflow now retains the bounded, accounted prefix and its indexes instead of discarding successful compile work. Paged schedule discovery reuses complete cached sources and derives only the uncovered tail exactly, so dense growing-range workbooks avoid rebuilding the same overflowed topology on every recalculation.
 - Experimental authoritative FormulaPlane evaluation now uses the existing consumer-read interval index when exact schedule construction falls back after topology-cache overflow, avoiding full-table scans per dirty formula during the first evaluation of bulk-loaded workbooks. (#240)
 - Experimental authoritative FormulaPlane evaluation now caches accounted consumer and precedent topology across warm and value-only evaluations. Exact graph, authority, semantic, provider, and dynamic-reference revisions invalidate or bypass stale cache generations; candidate, edge, and retained-byte cache overflow selects exact paged/run/native/repeated-pass request topology without span demotion, while span-free and warm no-dirty requests retain topology-free sparse paths.
--- a/crates/formualizer-eval/src/engine/eval.rs
+++ b/crates/formualizer-eval/src/engine/eval.rs
@@ -28,9 +28,10 @@ use crate::engine::{
     EvaluationRequestOutcome, EvaluationResourceBaselineStats, EvaluationResourceReason,
     EvaluationResourceRequestStats, FormulaDirtyLeaseOutcome, FormulaIngestBatch,
     FormulaIngestRecord, FormulaIngestReport, FormulaParseDiagnostic, FormulaParsePolicy,
-    FormulaPlaneMode, FormulaPlaneTopologyCacheOutcome, FormulaPlaneTopologyStrategy,
-    ResourceLedger, RowVisibilitySource, ScheduleUnit, Scheduler, VertexId, VertexKind,
-    VisibilityMaskMode,
+    FormulaPlaneMode, FormulaPlaneRoute, FormulaPlaneRouteEvent, FormulaPlaneRoutePhase,
+    FormulaPlaneRouteTransitionReason, FormulaPlaneTopologyCacheOutcome,
+    FormulaPlaneTopologyStrategy, ResourceLedger, RowVisibilitySource, ScheduleUnit, Scheduler,
+    VertexId, VertexKind, VisibilityMaskMode,
 };
 use crate::formula_plane::placement::prepare_anchor_once_fragment;
 use crate::formula_plane::placement::{
@@ -473,12 +474,31 @@ pub(crate) fn classify_mixed_topology_incomplete(
     }
 }
 
+#[derive(Clone, Debug, Default)]
+struct LegacyIslandPlan {
+    membership: Vec<VertexId>,
+    dirty_vertices: Vec<VertexId>,
+    sheet_ids: Vec<SheetId>,
+    island_id: u64,
+    retained_bytes: usize,
+    boundary_relationships: usize,
+    omitted_relationships: usize,
+    omitted_direct_relationships: usize,
+}
+
+impl LegacyIslandPlan {
+    fn is_empty(&self) -> bool {
+        self.membership.is_empty()
+    }
+}
+
 type FormulaPlaneMixedScheduleBuild = (
     MixedSchedule,
     BTreeMap<crate::formula_plane::runtime::FormulaSpanId, FormulaSpanRef>,
     u64,
     Vec<VertexId>,
     Vec<usize>,
+    LegacyIslandPlan,
 );
 
 #[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
@@ -901,6 +921,9 @@ pub struct Engine<R> {
     pub recalc_epoch: u64,
     snapshot_id: std::sync::atomic::AtomicU64,
     topology_epoch: u64,
+    /// False after a structural axis operation. While #171 remains open we
+    /// must not use relocated span read summaries to prove disconnection.
+    legacy_island_structural_summaries_trusted: bool,
     cached_static_schedule: Option<CachedScheduleEntry>,
     cached_mixed_topology: Option<CachedMixedTopology>,
     mixed_topology_cache_builds: u64,
@@ -1029,6 +1052,8 @@ pub struct Engine<R> {
 
     #[cfg(test)]
     last_formula_plane_span_eval_report: Option<SpanEvalReport>,
+    #[cfg(test)]
+    evaluation_request_begin_count_for_test: u64,
     before_prepared_span_commit_hook: Option<Box<dyn FnOnce() + Send + Sync>>,
     before_target_preparation_commit_hook: Option<Box<dyn FnOnce() + Send + Sync>>,
     #[cfg(test)]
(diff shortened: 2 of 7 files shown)


PSU3D0:
Reviewed independently and merging. Verified measurements, both mutations, and the soundness argument against blocker 1.

## The isolation proof is sound, and I want to record why

`prove_sheet_disjoint_legacy_island` (`eval.rs:18876`) fails closed on untrusted structural summaries (the #171 guard), any workbook- or sheet-scoped name, any spill, and any dynamic or array vertex. It then computes span boundary sheets as the union of sheets a span **writes** (`span_result_sheets`) and sheets a span **reads** (`span_read_sheets`), and rejects an island if the vertex's own sheet, any direct dependency's sheet, or any range dependency's sheet is in that set.

That covers both cycle directions against blocker 1:

- legacy depends on a span → the span's result sheet is in `span_result_sheets`, so the dependency check rejects.
- a span depends on legacy → the span reads that sheet, so it is in `span_read_sheets`, and the vertex's own-sheet check rejects.

So a contracted island cannot be span-connected, and the mixed schedule remains the sole detector of cycles routing through spans, exactly as "Gotcha G8" requires. `post_warm_cross_sheet_back_edge_demotes_before_legacy_tarjan` pins the live case: a post-warm `Aux!A5 = Sheet1!B5` back-edge through a span member with no graph vertex, asserting `cycle_errors == 1`, both cells `#CIRC`, an unrelated column still computing, and the span count dropping 2 → 1.

## Measurements reproduced

Freshly built release binary, verified rebuilt:

**The target** — three-section mixture at N=1000, `main` was 619 ms: **76 / 74 / 110 ms**, mismatches 0. That is the sum-of-its-parts result the issue asked for.

**Full corpus**, no regression, coverage and correctness steady:

| N | `Off` | auth first | warm hit | coverage | spans | mismatches |
|---|---|---|---|---|---|---|
| 500 | 86 ms | 518 ms | 0 ms | 69.2% | 9 | 0 |
| 1000 | 176 ms | 719 ms | 1 ms | 69.2% | 9 | 0 |
| 2000 | 577 ms | 3,185 ms | 3 ms | 69.2% | 9 | 0 |

`fmt`, `clippy -D warnings`, 30/30 soak and all five CI jobs green.

## A correction I owe you

I initially reported both targeted mutations as surviving, and concluded the optimization was unpinned. **That was my error, twice.** There are two contraction paths — `prove_sheet_disjoint_legacy_island` and, on fallthrough, `contract_legacy_islands` (`eval.rs:19377`). I disabled only the first, so contraction still happened via the second and nothing failed.

Disabling **both** produces the correct result: `mixed_tail_reads_complete_in_one_authoritative_pass` panics with its own message, "disabling island contraction must fail this routing assertion", and `independent_iterative_island_preserves_accumulator_and_single_request_lifecycle` also fails. The optimization is genuinely pinned, and the route-event telemetry is doing the job it was added for. Your claim was right and my verification was wrong.

I also briefly mis-read `prove_sheet_disjoint_legacy_island` as reintroducing the rejected per-sheet routing on the strength of its name. It does not — sheet disjointness is used as a conservative sufficient condition for isolation, not as a routing unit, and it falls through to the general path rather than replacing it. The name is the only thing about it that invites that misreading.

## Worth noting for whoever picks this up next

The two-path structure is load-bearing but not obvious, and it is exactly what made a plausible-looking mutation give a false negative. A short comment at the fallthrough in `contract_legacy_islands` noting that it is the general path behind the sheet-disjoint fast path would save the next reviewer the same detour.

PR #253, opened 2026-07-30
test(eval): pin topology retention after candidate overflow

### LINKED ISSUE #250: Topology-prefix retention is not pinned by a test: overflow-then-Hit needs engine-level coverage
Small test gap left by #249, filed so it is tracked rather than remembered.

## What is unprotected

#249 made candidate/edge overflow retain an accounted, bounded topology prefix instead of discarding the compile, which flips the next request from `SkippedOverflow` to `Hit` and skips the recompile. Measured effect on the `fp_coverage` corpus, against a control running `main`'s engine with the same probe: edited recalculation at 26k cells went from **1,818 ms to 3 ms**, and at 13k cells from **701 ms to 1 ms**.

That headline behaviour is not pinned by a test. Two mutations demonstrate the gap:

- Forcing `reuse_partial_topology = false` in `Engine`'s topology path: **all 2,447 tests still pass** and the measured win is unchanged.
- Making `schedule_dirty_work_paged_hybrid` ignore the retained prefix: `candidate_cap_retains_a_safe_prefix_and_exactly_schedules_the_tail` correctly fails, so prefix *consumption* is covered — but `warm_hit` at N=2000 stays at 4 ms rather than regressing to `main`'s 1,818 ms.

So the timing win comes from **retaining the compile at all**, not from consuming the prefix contents, and nothing asserts the retention. A change that stopped retaining on overflow would keep CI green while silently restoring a per-request recompile — which is precisely how the original `O(P x R)` scan survived twelve days in `main` (see the provenance note on #240: it was a known deferral recorded only in a PR body).

## Suggested test

An engine-level test that:

1. builds a workbook large enough to exceed `max_formula_plane_cache_candidates`,
2. evaluates once and asserts the first request reports candidate overflow,
3. makes a small value edit and evaluates again,
4. asserts the second request's topology cache outcome is `Hit` rather than `SkippedOverflow`.

Assert on the telemetry (`last_evaluation_resource_request_stats().topology.cache_outcome`) rather than on timing, so it is deterministic in CI. `probe-fp-coverage` already reads exactly these fields after #249, so the accessor path is established.

Worth also asserting the reverse direction stays intact: zero-limit, no-useful-prefix and retained-byte overflow must still skip, and a perpetual skip must still preserve values and increment the skip streak.

PSU3D0:
## What
- Add an engine-level authoritative-mode regression workbook with three promoted formula spans and a candidate cap of one.
- Assert the first evaluation observes two candidates against that cap, retains the one-edge useful prefix as `Built`, and routes through exact paged scheduling.
- Edit one input value and assert the second evaluation is a telemetry `Hit`, performs no topology compile work, and preserves hardcoded edited and untouched values.
- Keep production code unchanged and reuse existing zero/no-prefix, retained-byte-overflow, and perpetual-skip coverage for the negative direction.

## Validation
- `cargo test -p formualizer-eval --lib --all-features candidate_overflow_retains_topology_for_next_request_hit -- --nocapture` (1 passed)
- `cargo test -p formualizer-eval --lib --all-features engine::tests::evaluation_resource_observability -- --nocapture` (8 passed)
- `cargo test -p formualizer-eval --lib --all-features skipped_topology_is_typed_atomic_and_never_cached -- --nocapture` (1 passed)
- `cargo fmt --all --check`
- `cargo test -p formualizer-eval --lib --all-features` (2,450 passed, 12 ignored)

## Mutation evidence
- Disabled overflow compile retention itself by changing the partial-cache branch from `else if !relationships.is_empty() && matches!(...)` to `else if false && !relationships.is_empty() && matches!(...)` in `compile_mixed_topology`.
- A freshly rebuilt baseline test binary hashed `189d98f833d3b696b1ef6649713846ac4e62ae62c8e27edc35904543d8423f00`; the rebuilt mutant binary timestamp advanced and hashed `35a4ba6237828dadeaa9da5729d04c0d4f5f3b494aff829a5fe2b6acfafdf2ff`.
- `candidate_overflow_retains_topology_for_next_request_hit` failed as required: telemetry was `SkippedOverflow` instead of expected `Built`.
- Restored the implementation and freshly rebuilt the passing binary before final validation.

Closes #250

### DIFF
--- a/crates/formualizer-eval/src/engine/tests/evaluation_resource_observability.rs
+++ b/crates/formualizer-eval/src/engine/tests/evaluation_resource_observability.rs
@@ -259,6 +259,83 @@ fn topology_build_hit_and_overflow_materialization_are_exactly_observed() {
 }
 
 #[test]
+fn candidate_overflow_retains_topology_for_next_request_hit() {
+    let mut config =
+        EvalConfig::default().with_formula_plane_mode(FormulaPlaneMode::AuthoritativeExperimental);
+    config.max_formula_plane_cache_candidates = 1;
+    let mut engine = Engine::new(TestWorkbook::default(), config);
+    let mut formulas = Vec::new();
+    for row in 1..=100 {
+        engine
+            .set_cell_value("Sheet1", row, 1, LiteralValue::Number(row as f64))
+            .unwrap();
+        formulas.push(record(&mut engine, row, 2, &format!("=A{row}*2")));
+        formulas.push(record(&mut engine, row, 3, &format!("=B{row}+1")));
+        formulas.push(record(&mut engine, row, 4, &format!("=C{row}+1")));
+    }
+    engine
+        .ingest_formula_batches(vec![FormulaIngestBatch::new("Sheet1", formulas)])
+        .unwrap();
+    assert_eq!(engine.baseline_stats().formula_plane_active_span_count, 3);
+
+    engine.evaluate_all().unwrap();
+    let overflow = engine.last_evaluation_resource_request_stats().unwrap();
+    assert_eq!(
+        overflow.topology.cache_outcome,
+        FormulaPlaneTopologyCacheOutcome::Built
+    );
+    assert_eq!(
+        overflow.topology.strategy,
+        FormulaPlaneTopologyStrategy::ExactPagedIndexed
+    );
+    assert_eq!(overflow.topology.candidate_cap, Some(1));
+    assert_eq!(
+        overflow.topology.overflow_reason,
+        Some(EvaluationResourceReason::FormulaPlaneTopologyCandidates)
+    );
+    assert_eq!(overflow.topology.candidates_observed, 2);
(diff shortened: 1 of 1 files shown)


PSU3D0:
Independent review complete. The test is non-vacuous and pins the intended layer:

- The fixture creates three promoted spans, sets candidate cap `1`, and observes exactly two candidates plus one retained relationship.
- The first request reports candidate overflow but `Built`, proving a useful partial topology was retained rather than skipped.
- The edited second request reports `Hit` with zero build events and zero observed compile work, while hardcoded edited and untouched values remain correct.
- Disabling the actual incomplete-topology retention branch changes the first outcome to `SkippedOverflow` and fails the named test. This distinguishes compile retention from partial-prefix consumption.
- Existing tests already cover zero/no-useful-prefix, retained-byte overflow, perpetual skip streak, and exact tail scheduling; duplicating those here would add no protection.

Two fresh adversarial reviewers found no fixes worth doing. Local focused validation and all five CI jobs pass.

Issue #233, opened 2026-07-25
FunctionContext::cancellation_token still returns a raw flag after the CancelToken unification


PSU3D0:
`FunctionContext::cancellation_token()` still returns `Option<Arc<AtomicBool>>`, while every engine, workbook and SheetPort entry point now takes a `CancelToken` after #232. The public API therefore carries two cancellation types.

This was left deliberately rather than rushed into #232. The trait method is a *read* interface for custom-function authors, and converting it cascades: `SpanEvaluator::new_with_cancel`, the aggregate range walks, and several other internals take `&AtomicBool` and would all need threading in the same change. `CancelToken::as_flag` and `CancelToken::from_flag` bridge the two cleanly in the meantime, so nothing is broken -- it is an inconsistency, not a defect.

Worth settling before `0.8.0`, because after a release the trait signature is frozen and custom-function authors will have written against it.

Two options:

1. Change the trait to `fn cancellation_token(&self) -> Option<CancelToken>` and thread the token through the span evaluator and range-walk helpers. Consistent, and gives function authors the same ergonomics as callers.
2. Keep the raw flag on the trait deliberately and document why: function authors poll in tight loops, where an `Arc<AtomicBool>` is the cheapest possible check and a newtype adds nothing.

Option 1 is the tidier end state; option 2 is defensible if the polling-cost argument holds. Either is fine, but the current split should be a decision rather than an accident.

drafted with love by (agent) Manfred, with approval from Frankie

PSU3D0:
## Pre-0.8.0 decision: migrate the full public cancellation seam

Choosing the consistent option before the API freeze:

- `EvaluationContext::cancellation_token()` → `Option<CancelToken>`
- `FunctionContext::cancellation_token()` → `Option<CancelToken>`
- public `RangeView::with_cancel_token(...)` and its stored token → `CancelToken`
- private `SpanEvaluator` borrows/polls `CancelToken`
- private engine routines may continue bridging through `as_flag()` in this change

Changing only `FunctionContext` would not close the representation leak: `DefaultFunctionContext` forwards from the also-public `EvaluationContext`, and `RangeView::with_cancel_token` is a third public raw-flag seam. All three conflict with `CancelToken`'s documented promise that its underlying representation stays private so richer cancellation can be added without a later break.

[...]

PR #256, opened 2026-07-30
ci: enforce public Rust API snapshots

### LINKED ISSUE #231: Add a public-API surface check to CI
The pre-0.8.0 API review found test-only and benchmark-only affordances on the published surface -- a `pub` commit hook whose sibling was already `pub(crate)`, and a SheetPort accessor used only by an unpublished benchmark crate. None of that was caught in review, because a change to the public surface looks like any other diff.

The review itself had to be reconstructed after the fact from `git diff v0.7.1..main`, filtering 287 added `pub` items by reachability. That is not a repeatable process and it does not scale to every PR.

Proposal: add a public-API snapshot to CI, either `cargo public-api` (requires a nightly toolchain for rustdoc JSON) or an equivalent committed surface file. The check should fail when the surface changes without the snapshot being updated in the same PR, so that adding, removing, or widening public API becomes an explicit, reviewable act.

Scope worth deciding: which crates to cover. `formualizer-eval` and `formualizer-common` carry most of the surface and would deliver most of the value; `formualizer-sheetport`, `formualizer-workbook` and `sheetport-spec` are also published and would benefit.

This matters more after `0.8.0` than before it, since every subsequent change to a released surface is a semver event.

PSU3D0:
Closes #231

## Scope

This commits deterministic `cargo-public-api` snapshots for the eight crates.io/release surfaces: `formualizer`, `formualizer-common`, `formualizer-parse`, `formualizer-eval`, `formualizer-workbook`, `formualizer-sheetport`, `sheetport-spec`, and `formualizer-macros`.

It deliberately excludes CFFI (installed C header and exported ABI are the contract), Python (generated `.pyi` files are already diffed), WASM/npm (wasm-bindgen/TypeScript package output is the contract), and unpublished bench/testkit/xtask crates. No manifest publish flags change here; the CFFI/Python/WASM flags remain release-hygiene debt for the final API-review disposition.

`scripts/public-api.sh` owns explicit maximal native feature allowlists, excludes `benchmark_internal` and JS/browser profiles, never uses `--all-features`, generates all requested outputs before transactionally publishing any snapshot, and supports setup/check/update with optional validated crate names. CI adds a separate `public-api` job without changing the normal Rust job.

## Determinism

The generator pins exact `cargo-public-api =0.52.0` in a versioned repo-local tool root, installer Rust `1.93.0`, rustdoc `nightly-2026-02-16`, and target `x86_64-unknown-linux-gnu`. It sets `LC_ALL=C`, `TZ=UTC`, and `CARGO_TERM_COLOR=never`; neutralizes plain/encoded flags, compiler overrides, wrappers, linker/runner overrides, and Cargo build equivalents; and isolates both `CARGO_TARGET_DIR` and `CARGO_HOME`.

Snapshots omit blanket and compiler auto-trait implementations while retaining explicit and derive-generated implementations. The README calls out the resulting Send/Sync limitation.

Two consecutive clean full generations/diffs passed with the same committed baseline (aggregate snapshot SHA-256 evidence: `678eb4a2611bab968b5b7f8bcb83c8b0447d16542c9f0d956b20eb3fd6cdfabf`). The post-#233 `CancelToken` signatures are present in the baseline.

## Baseline size

The eight snapshots total 35,445 lines and 4,469,997 bytes:

| Crate | Lines | Bytes |
| --- | ---: | ---: |
| `formualizer` | 59 | 2,195 |
| `formualizer-common` | 1,423 | 116,116 |
| `formualizer-parse` | 1,030 | 96,015 |
| `formualizer-eval` | 29,393 | 3,857,614 |
| `formualizer-workbook` | 2,487 | 307,228 |
| `formualizer-sheetport` | 379 | 34,373 |
| `sheetport-spec` | 672 | 56,381 |
| `formualizer-macros` | 2 | 75 |

The broad existing eval surface is mechanically baselined and remains an input to the final pre-tag API review. A comparison with auto-derived impls omitted still totaled 3,803,937 bytes, so retaining downstream `Clone`/`Eq`/`Default` commitments is preferable; future drift remains localized.

## Verification

- `scripts/public-api.sh setup`
- `scripts/public-api.sh check` (full baseline, twice consecutively, and final post-mutation pass)
- focused check with hostile inherited Rust/Cargo environment variables
- `cargo fmt --all -- --check`
- `bash -n scripts/public-api.sh` (`shellcheck` was unavailable)
- unknown-crate rejection returned 2 with usage
- no-partial-update probe failed on the second requested crate and left both baseline hashes unchanged
- workflow YAML parsed with Ruby Psych (`actionlint` was unavailable)
- hostile absolute/apostrophe `TMPDIR`, `CDPATH`, encoded cfg flags, compiler, wrapper, and target overrides were neutralized; relative `TMPDIR` was safely rejected
- snapshot destination, intermediate target, and unexpected proc-macro symlinks were rejected before mutation
- injected publication failure after the first selected install rolled back every selected snapshot byte-for-byte
- temporary unclassified manifest feature failed before generation with actionable classification instructions
- injected `diff` status 2 returned an operational error without snapshot update instructions
- a fake global `cargo-public-api` on `PATH` was not invoked; the exact isolated binary was used
- inherited `CARGO=/bin/false` was neutralized and the focused `formualizer-macros` check passed
- deterministic two-writer regression test paused update A after its first install, proved update B waited on the exclusive lock, and finished with the complete B/B candidate set rather than A/B mixing
- concurrency harness uses EXIT-only cleanup with INT/TERM handlers exiting 130/143; an injected TERM returned 143, restored all four mutated files, and removed its temporary directory exactly once
- two simultaneous focused checks both acquired shared locks and passed
- a symlinked lock file was rejected before opening or generation

Mutation verification used the exact script path CI runs. Adding `pub fn __public_api_ci_probe() {}` to `formualizer-eval` made `check formualizer-eval` fail with the added line. Updating that probed snapshot, restoring source, and checking again failed with the removed line. A trap restored source and baseline; the final full check passed with a clean tree.

## Review and updates

This detects drift, not semver intent. Reviewers should classify snapshot changes, separately assess inferred Send/Sync changes, and treat the unexpectedly broad existing public surface and binding manifest publish flags in the final API review.

For an intentional change, run `scripts/public-api.sh setup`, then `scripts/public-api.sh update [crate ...]`, review the unified diff, run `scripts/public-api.sh check`, and commit the snapshot with the source change.

## Hardening follow-up

The follow-up commit derives a physical root with `CDPATH` neutralized, validates repository containment and rejects symlinked target/snapshot components, safely manages the proc-macro rustdoc bridge, and uses cleanup functions instead of interpolated trap data. Updates stage candidates on the snapshot filesystem, atomically move each file with backups, and roll back the complete selected set on ERR/INT/TERM. Checks now hold a shared Linux `flock` and updates an exclusive lock from before generation through compare/publication/rollback, preventing cooperating processes from observing or producing an interleaved baseline.

Every governed package's `cargo metadata --no-deps` feature keys must now be classified as enabled native, explicitly excluded, alias, or default. New features cannot silently escape profile review. Python 3 and the Linux/devcontainer canonical-generation requirement are documented.

The new job runs on `ubuntu-24.04` with `contents: read`; checkout, cache, and both rust-toolchain invocations are pinned to official full commit SHAs resolved from their current v4/stable/nightly refs. The isolated tool root, Cargo home, and build target are the only cached paths. CI also runs the deterministic concurrent-update regression test. Hostile non-cooperating filesystem mutation, SIGKILL, process crashes that bypass traps, and power loss are explicitly outside the rollback guarantee; detected rollback failures retain recoverable backup directories.


PSU3D0:
Final review complete. Merging this as a release gate, not merely a generated-file dump.

The initial baseline is intentionally large—35,445 lines / 4,469,997 bytes, with 29,393 lines in `formualizer-eval`. Comparative generation proved this is primarily real reachable/macro-authored surface: omitting all derived impls still left 3.80MB, so losing downstream `Clone`/`Eq`/`Default` commitments would not buy meaningful reviewability. The baseline is mechanically reproduced; future drift diffs are localized. Its breadth is now explicit input to the final pre-tag API review.

Adversarial review materially hardened the gate before merge:

- exact isolated tooling and immutable CI action SHAs
- canonical OS/toolchain/target/features with metadata-enforced feature classification
- hostile `TMPDIR`, `CDPATH`, `CARGO`, encoded flags, compiler/wrapper and linker overrides neutralized or rejected
- symlink/containment guards around target, lock and snapshot paths
- operational `diff` failures distinguished from API drift
- staged multi-file publication with rollback on detected errors/signals
- shared/exclusive `flock` preventing checks or concurrent updates from observing/producing an interleaved baseline
- deterministic regression harness proving two writers finish with one complete candidate set, plus signal-safe exactly-once restoration

A reviewer suggested replacing the now-large shell safety machinery with straightforward generate/copy. I am rejecting that suggestion: generate/copy is the design under which partial publication and concurrent A/B mixing were both reproduced. The complexity is paid for by focused failure, mutation and concurrency tests and is preferable to a simpler gate that can corrupt its own authority.

All eight crate snapshots reproduce, add/remove public-item mutations fail in both directions, unclassified features fail, rollback and concurrency mutations pass, and all six CI jobs—including the new public-api job—are green.

Issue #320, opened 2026-08-11
[bug] eval: MATCH with match_type -1 returns the wrong row on descending data of 8 or more entries


tommy230:
### Summary

The `match_type = -1` branch of `binary_search_match` selects the **first**
entry that is `>=` the needle. On correctly descending data the answer is the
**last** such entry. The `< 8` linear path in `MatchFn` gets this right, so
the defect only appears once the vector reaches 8 entries, and is masked
whenever the needle hits a key exactly (the search returns early on
equality).

### Reproduction

1. `A1:A10` = `10, 9, 8, 7, 6, 5, 4, 3, 2, 1`.
2. Evaluate `=MATCH(3.5,A1:A10,-1)`.

### Minimal input

| Formula | Excel | Formualizer 0.8.1 |
| --- | --- | --- |
| `=MATCH(3.5,A1:A10,-1)` | 7 | `1` |
| `=MATCH(3,A1:A10,-1)` (exact hit, masked) | 8 | 8 |

### Expected / Actual

Expected 7 (value `4`, the smallest entry `>= 3.5`); got 1 (value `10`).

### Additional context

`crates/formualizer-eval/src/builtins/lookup/core.rs`, the `else` branch of
`binary_search_match`: `if c >= 0 && best.is_none_or(|b| i < b)`. Verified
present on unmodified `main` at `fe7b9111`. Not fixed in the Issue B PR,
which changes only *which entries* that branch visits, not how it picks
among them.

### Environment

- Formualizer version: 0.8.1 (`main` at `fe7b9111`)
- Host: Rust
- OS: macOS 15 (Darwin 25.5.0), aarch64
- Compiler: rustc 1.93.0 (CI-pinned toolchain)
- Enabled features: `--all-features`

Oracle for every expected value: Microsoft Excel 16.105.3 (Microsoft 365 for Mac), measured on throwaway synthetic workbooks.

Issue #319, opened 2026-08-11
[bug] eval: MATCH exact-match coerces a blank cell to 0, so MATCH(0, range, 0) matches an empty row


tommy230:
### Summary

Exact `MATCH` finds a blank cell when the lookup value is `0`. Excel returns
`#N/A`: a blank never equals `0` in an exact match.

### Reproduction

1. `A1:A5` = `1, 2, 3, 4, 5`; leave `A6:A10` unwritten.
2. Evaluate `=MATCH(0,A1:A10,0)`.

### Minimal input

| Formula | Excel | Formualizer 0.8.1 |
| --- | --- | --- |
| `=MATCH(0,A1:A10,0)` | `#N/A` | `6` |
| `=MATCH(3,A1:A10,0)` (control) | 3 | 3 |

### Expected / Actual

Expected `#N/A`; got `6`, the first blank row. `value_to_f64_lenient` maps
`Empty` to `0.0` and the exact path compares through it.

### Additional context

Found while fixing Issue B; it lives on the exact-match path, which that fix
does not touch. Worth deciding deliberately, because `Empty -> 0.0` is load
bearing elsewhere (see #279, #285) and the right scope may be the coercion
helper rather than the lookup call sites.

### Environment

- Formualizer version: 0.8.1 (`main` at `fe7b9111`)
- Host: Rust
- OS: macOS 15 (Darwin 25.5.0), aarch64
- Compiler: rustc 1.93.0 (CI-pinned toolchain)
- Enabled features: `--all-features`

Oracle for every expected value: Microsoft Excel 16.105.3 (Microsoft 365 for Mac), measured on throwaway synthetic workbooks.

PSU3D0:
Deferred out of 0.8.2 deliberately, and worth recording why, since your instinct in the original report was right.

`Empty -> 0.0` in `value_to_f64_lenient` is load bearing across at least #279 (blank cells reached through formulas publish numeric 0) and #285 (COUNTBLANK/COUNTIF over never-written blanks). Changing it at the coercion helper moves behaviour in criteria matching and aggregation at the same time as lookups, and changing it only at the lookup call sites leaves the helper meaning two different things depending on caller. Neither is a patch-release change.

[...]

Issue #285, opened 2026-08-06
[bug] eval: COUNTBLANK/COUNTIF over never-written blank ranges deviate from Excel (possible sparse-range artifact)


PSU3D0:
Three blank-range aggregation probes deviate from Excel, reproduced identically on `95b67049` and on the #279 fix branch (so pre-existing and unrelated to result finalization), all with unset cells `C1:C3` under the engine test harness:

`=COUNTBLANK(C1:C3)` returns 0 where Excel returns 3. `=COUNTIF(C1:C3,"")` returns 0 where Excel returns 3. `=SUM(IF(C1:C2="",1,0))` returns `#VALUE!` where Excel returns 2.

Found during the #279 adversarial review. The reviewer noted this may be a sparse-range artifact of the test harness rather than engine semantics — never-written cells may not materialize in range views the way saved-workbook blanks do — so the first step is discrimination: reproduce via the workbook API and via a loaded xlsx with genuinely empty cells inside a used range, and compare against LibreOffice for each surface. If the deviation only exists in the test harness, the fix is harness-side and the parity implications are nil; if it reproduces on real workbooks, COUNTBLANK/COUNTIF-blank semantics are a user-visible parity gap that the Enron parity runner would likely surface at frequency.

Either way this should be resolved or explicitly triaged before the 0.8.0 parity baseline is published, so the baseline number is not polluted by a known-uninvestigated class.

drafted with love by (agent) Manfred, with approval from Frankie

Issue #279, opened 2026-08-06
[bug] eval: blank-cell formula results escape as empty instead of numeric zero, propagating false blankness downstream


PSU3D0:
Formulas whose final evaluated result is a blank-cell passthrough return an empty value instead of Excel's numeric zero, and the blankness incorrectly propagates to downstream cells.

With `C1` truly blank, Formualizer at `95b67049` returns empty (`None`) for `=C1`, `=+C1`, `=IF(TRUE,C1,5)`, `=IFERROR(C1,5)`, and `=CHOOSE(1,C1)`. Excel and LibreOffice (24.2.7 headless recalc, verified today) return numeric `0` for all five.

The downstream evidence shows conversion must happen at the producing cell, not at display: with `A17` containing `=C1`, LibreOffice gives `ISBLANK(A17)` = FALSE, `COUNT(A17)` = 1, and `COUNTA(A17)` = 1; Formualizer gives TRUE, 0, and 0 because the raw `Empty` escapes as the cell's computed value and downstream consumers treat the formula cell as blank.

Everything else already matches the oracle and must not change: `ISBLANK(C1)` = TRUE, `=C1&"x"` = `"x"`, `=LEN(C1)` = 0, `=C1=0` = TRUE, `=C1=""` = TRUE, `=T(C1)` = `""`, `=N(C1)` = 0, `=ISNUMBER(C1)` = FALSE, `=COUNT(C1)` = 0, `=COUNTA(C1)` = 0, `=TEXT(C1,"0")` = `"0"`. In particular, direct-reference inspection semantics (`ISBLANK`) and range/criteria semantics over genuinely blank cells are correct today and are not in scope.

Scope: convert a formula's final `Empty` result to numeric zero at result finalization, leaving intra-expression `Empty` coercion, stored blank cells, and reference inspection untouched. This changes computed values surfaced through persistence and bindings for affected cells (0 instead of empty), which should be called out in the changelog.

Found while preparing the 0.8.0 parity pass; oracle probes are recorded in the platform workspace.

drafted with love by (agent) Manfred, with approval from Frankie

PR #321, opened 2026-08-11
fix(eval): temporal values compare as Excel serials in lookups (#316)

### LINKED ISSUE #316: [bug] eval: temporal values are incomparable in lookups, so every lookup over a date column returns #N/A
### Summary

`MATCH`, `VLOOKUP` and `HLOOKUP` cannot find or order a `Date`, `DateTime`
or `Time` value. Exact lookups miss, and approximate lookups fail their
sortedness guard and return `#N/A`, for data that Excel resolves.

### Reproduction

1. Write date-typed values into `A1:A5` (2024-01-01 .. 2024-01-05) — an
   ingested date-formatted cell, or a `date` written through the API. A
   formula-produced `=DATE(...)` yields a plain number and does not
   reproduce.
2. Write `10, 20, 30, 40, 50` into `B1:B5`.
3. Write the date 2024-01-03 into `D1`.
4. Evaluate the formulas below.

### Minimal input

| --- | --- | --- |
| `=MATCH(D1,A1:A5,0)` | 3 | `#N/A` |
| `=MATCH(D1,C1:C5,0)` where `C` holds the same dates as plain serials | 3 | `#N/A` |
| `=MATCH(DATE(2024,1,3),A1:A5,1)` | 3 | `#N/A` |
| `=MATCH(DATE(2024,1,4)+0.5,A1:A5,1)` | 4 | `#N/A` |
| `=VLOOKUP(D1,A1:B5,2,FALSE)` | 30 | `#N/A` |
| `=VLOOKUP(DATE(2024,1,3),A1:B5,2,TRUE)` | 30 | `#N/A` |

### Expected behavior

Excel has no date type at the formula level: a date cell holds a serial
carrying a date number format. A temporal value must therefore order and
match exactly as the number it is, against other temporal values and against
plain numerics alike.

### Actual behavior

Three places treat the temporal variants as unusable:

- `builtins::lookup::lookup_utils::value_to_f64_lenient` returns `None` for
  them, so `cmp_for_lookup` reports them incomparable and the approximate
  sortedness guard rejects any date column.
- `find_exact_index_in_view` has no temporal arm and falls through to
  `Ok(None)`.
- `LookupHashKey::from_literal` returns `None` for them, so the hash index
  can key neither a date needle nor a date cell.

### Additional context

Not a 0.8.x regression; the same code is present at `0db29966` and earlier.
The trigger requires a genuine date-typed value, which is why the JSON
formula suite does not catch it. Fix and regression tests ready — will open
a PR against this issue.

### Environment

- Host: Rust

tommy230:
Fixes #316.

## What was wrong

`MATCH`, `VLOOKUP` and `HLOOKUP` could not find or order a `Date`,
`DateTime` or `Time` value. Three independent places refuse the temporal
variants:

- `builtins::lookup::lookup_utils::value_to_f64_lenient` returns `None` for
  them. `cmp_for_lookup` is built on it, so every temporal comparison is
  "incomparable" — which means `MATCH`'s approximate sortedness guard sees an
  unordered vector and returns `#N/A` for a perfectly ordered date column.
- `find_exact_index_in_view` dispatches on the needle's variant and has no
  temporal arm, so a date-typed needle falls through to `Ok(None)`.
- `LookupHashKey::from_literal` returns `None` for them, so the lookup hash
  index can key neither a date needle nor a date cell and the fast exact path
  is unusable.

## Excel behaviour

Excel has no date type at the formula level. A date cell holds a serial and a
date number format; the format is display, not type. So a lookup over dates
is a lookup over numbers, and a date must order against another date, against
a plain serial, and against a fractional serial, exactly as two numbers do.

Measured on Microsoft Excel 16.105.3 (Microsoft 365 for Mac), synthetic
workbook, `A1:A5` = 2024-01-01..05 as dates, `B1:B5` = 10..50, `C1:C5` = the
same dates as plain serials, `D1` = the date 2024-01-03:

| Formula | Excel | main @ `fe7b9111` |
| --- | --- | --- |
| `=MATCH(D1,A1:A5,0)` | 3 | `#N/A` |
| `=MATCH(D1,C1:C5,0)` | 3 | `#N/A` |
| `=MATCH(DATE(2024,1,3),A1:A5,1)` | 3 | `#N/A` |
| `=MATCH(DATE(2024,1,4)+0.5,A1:A5,1)` | 4 | `#N/A` |
| `=VLOOKUP(D1,A1:B5,2,FALSE)` | 30 | `#N/A` |
| `=VLOOKUP(DATE(2024,1,3),A1:B5,2,TRUE)` | 30 | `#N/A` |

The fractional-serial row is the interesting one: it can only come out at 4
if a `Date` and a non-integral number sit on the same number line.

## The fix

All three sites route through `LiteralValue::as_serial_number()`, which
already exists and already implements the 1900-system conversion the rest of
the crate uses.

Nothing that previously matched changes. `as_serial_number` is total over the
three temporal variants, so no case that used to return a position now
returns `#N/A`; the change only converts `#N/A` into the value Excel
produces. Non-temporal values take exactly the paths they took before.

## Scope

Not a regression: the same code is present well before 0.8.0. The trigger
requires a genuine date-typed value — an ingested date-formatted cell, or a
`date` written through the API. A formula-produced `=DATE(2024,1,3)` yields a
plain number and never reaches any of the three sites, which is why the JSON
formula suite does not catch this and why it shows up first on real ingested
workbooks. Any date-keyed schedule — rate tables, amortisation schedules,
period lookups — is affected wholesale rather than at the margin.

## Verification

Nine regression tests in
`crates/formualizer-eval/src/engine/tests/temporal_lookup_semantics.rs`, each
pinned to a measured Excel result. Revert the production change and **six of
the nine fail**. The three that survive are the controls:

- a plain serial needle against date cells (already worked; must keep working),
- a `Time` key found by its serial fraction via the numeric lane,
- an absent date still reporting `#N/A` — making dates comparable must not
  make every date match something.

Gates, run locally on the CI-pinned toolchain (macOS 15, aarch64):

```
cargo +1.93.0 fmt --all -- --check                                  exit 0
cargo +1.93.0 clippy --workspace --exclude formualizer-bench-core \
  --exclude formualizer-python --exclude formualizer-wasm \
  --all-targets --all-features -- -D warnings                       exit 0
cargo +1.93.0 test --workspace --exclude formualizer-bench-core \
  --exclude formualizer-python --exclude formualizer-wasm \
  --all-features --no-fail-fast            3471 passed, 1 failed
```

The single failure is pre-existing and unrelated: the JSON formula suite's
`IMEXP`/`IMSIN`/`IMCOS` cases differ in the last ULP on macOS arm64.
Unmodified `main` gives 3462 passed, 1 failed on the same host — same failure,
same three cases.

Not run here: the Linux-only `public-api` snapshot job. No public item is
added, removed or re-typed — `builtins::lookup::lookup_utils` does not appear
in `public-api/formualizer-eval.txt`, `LookupHashKey::from_literal` is
`pub(crate)`, and the `LookupHashKey` variants that *are* in the snapshot are
untouched.

- **Exact `MATCH` coerces blanks to zero** (#319): `=MATCH(0,A1:A10,0)`
  over a range with a blank tail returns 6 where Excel returns `#N/A`. Same
  helper (`Empty -> 0.0`), different path, and `Empty -> 0.0` is load-bearing
  elsewhere (#279, #285), so it deserves its own decision.
- **`MATCH` with `match_type = -1` on 8+ descending entries** (#320)
  returns the first entry `>=` the needle instead of the last. Unrelated to
  temporal values; verified present on unmodified `main`.
- Whether `Duration` should also compare as a serial. `as_serial_number`
  supports it, but `Duration` has no Excel counterpart and no lookup was
  observed to need it, so it stays incomparable.

Drafted by Claude (agent), verified by execution against a real Excel
oracle, reviewed by a human before submission.

### DIFF
--- a/crates/formualizer-eval/src/builtins/lookup/lookup_utils.rs
+++ b/crates/formualizer-eval/src/builtins/lookup/lookup_utils.rs
@@ -9,6 +9,10 @@ use formualizer_common::{ExcelError, ExcelErrorKind, LiteralValue};
 /// - Number / Int: numeric
 /// - Text: parsed if it looks numeric (lenient)
 /// - Boolean: TRUE=1, FALSE=0
+/// - Date / DateTime / Time: the Excel serial number. Excel has no date type
+///   at the formula level -- a date cell holds a serial carrying a date number
+///   format -- so a temporal value must order against plain numerics and
+///   against other temporal values exactly as two numbers do.
 /// - Empty: treated as 0
 pub fn value_to_f64_lenient(v: &LiteralValue) -> Option<f64> {
     match v {
@@ -16,6 +20,9 @@ pub fn value_to_f64_lenient(v: &LiteralValue) -> Option<f64> {
         LiteralValue::Int(i) => Some(*i as f64),
         LiteralValue::Text(s) => s.parse::<f64>().ok(),
         LiteralValue::Boolean(b) => Some(if *b { 1.0 } else { 0.0 }),
+        LiteralValue::Date(_) | LiteralValue::DateTime(_) | LiteralValue::Time(_) => {
+            v.as_serial_number()
+        }
         LiteralValue::Empty => Some(0.0),
         _ => None,
     }
@@ -354,6 +361,15 @@ pub fn find_exact_index_in_view(
         LiteralValue::Boolean(b) => find_exact_boolean_in_view(view, *b, vertical),
         LiteralValue::Empty => find_exact_empty_in_view(view, vertical),
         LiteralValue::Error(e) => Err(e.clone()),
+        // A temporal needle searches the numeric lane by serial: the arrow
+        // store keeps dates and times as serials under a temporal type tag,
+        // so an exact lookup for a date must not fall through to "no match".
+        LiteralValue::Date(_) | LiteralValue::DateTime(_) | LiteralValue::Time(_) => {
+            match needle.as_serial_number() {
+                Some(serial) => find_exact_number_in_view(view, serial, vertical),
+                None => Ok(None),
+            }
+        }
         _ => Ok(None),
     }
 }
--- a/crates/formualizer-eval/src/engine/lookup_index_cache.rs
+++ b/crates/formualizer-eval/src/engine/lookup_index_cache.rs
@@ -64,11 +64,14 @@ impl LookupHashKey {
             LiteralValue::Text(s) => Some(Self::Text(s.to_lowercase().into_boxed_str())),
             LiteralValue::Boolean(b) => Some(Self::Boolean(*b)),
             LiteralValue::Empty => Some(Self::Empty),
+            // Temporal values are numbers in Excel: key them by their serial so
+            // an exact lookup finds them whether the needle or the cell (or
+            // both) carry a temporal type rather than a plain numeric.
+            LiteralValue::Date(_) | LiteralValue::DateTime(_) | LiteralValue::Time(_) => value
+                .as_serial_number()
+                .map(|serial| Self::Number(normalize_f64_bits(serial))),
             LiteralValue::Error(_)
             | LiteralValue::Array(_)
-            | LiteralValue::Date(_)
-            | LiteralValue::DateTime(_)
-            | LiteralValue::Time(_)
             | LiteralValue::Duration(_)
             | LiteralValue::Pending => None,
         }
--- a/crates/formualizer-eval/src/engine/tests/mod.rs
+++ b/crates/formualizer-eval/src/engine/tests/mod.rs
@@ -187,3 +187,5 @@ mod scc_iterate;
 mod scc_runtime_cycles;
 mod scc_runtime_property;
 mod short_circuit_dispatch;
+
+mod temporal_lookup_semantics;
--- a/crates/formualizer-eval/src/engine/tests/temporal_lookup_semantics.rs
+++ b/crates/formualizer-eval/src/engine/tests/temporal_lookup_semantics.rs
@@ -1,0 +1,182 @@
+//! Temporal values participate in lookups as ordinary Excel serials.
+//!
+//! Oracle: Microsoft Excel 16.105.3 (Microsoft 365 for Mac), synthetic
+//! workbook, values recalculated in-app (`oracle: excel-verified`).
+//!
+//! Excel has no date type at the formula level: a date cell holds a serial
+//! carrying a date number format. Every lookup family member therefore
+//! compares dates against dates, and against plain numerics, exactly as it
+//! compares two numbers.
+
+use crate::engine::{Engine, EvalConfig};
+use crate::test_workbook::TestWorkbook;
+use chrono::{NaiveDate, NaiveTime};
+use formualizer_common::LiteralValue;
+use formualizer_parse::parser::parse;
+
+fn date(y: i32, m: u32, d: u32) -> LiteralValue {
+    LiteralValue::Date(NaiveDate::from_ymd_opt(y, m, d).unwrap())
+}
+
+/// Serial for 2024-01-01 in the Excel 1900 system.
+const JAN_1_2024: f64 = 45292.0;
+
+/// Column A: date-typed keys 2024-01-01..2024-01-05.
+/// Column B: numeric payload 10..50.
+/// Column C: the same keys written as plain numeric serials.
+/// D1: a date-typed needle cell (2024-01-03).
+fn build_date_engine() -> Engine<TestWorkbook> {
+    let mut engine = Engine::new(TestWorkbook::new(), EvalConfig::default());
+    for i in 1..=5u32 {
+        engine
+            .set_cell_value("Sheet1", i, 1, date(2024, 1, i))
+            .unwrap();
+        engine
+            .set_cell_value("Sheet1", i, 2, LiteralValue::Int((i as i64) * 10))
+            .unwrap();
+        engine
+            .set_cell_value(
+                "Sheet1",
+                i,
+                3,
+                LiteralValue::Number(JAN_1_2024 + (i as f64) - 1.0),
+            )
+            .unwrap();
+    }
+    engine
+        .set_cell_value("Sheet1", 1, 4, date(2024, 1, 3))
+        .unwrap();
+    engine
+}
+
+fn eval(engine: &mut Engine<TestWorkbook>, formula: &str) -> Option<LiteralValue> {
+    engine
+        .set_cell_formula("Sheet1", 1, 10, parse(formula).unwrap())
+        .unwrap();
+    engine.evaluate_all().unwrap();
+    engine.get_cell_value("Sheet1", 1, 10)
+}
+
+fn assert_number(value: Option<LiteralValue>, expected: f64, formula: &str) {
+    match value {
+        Some(LiteralValue::Int(i)) => assert_eq!(i as f64, expected, "{formula}"),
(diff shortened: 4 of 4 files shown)


PSU3D0:
Thank you for this — and for #316, which is a genuinely good report. I verified the defect independently against the published 0.8.1 wheel before looking at the patch: `=VLOOKUP(<date>,A1:B5,2,FALSE)` returns `#N/A` where Excel gives 30, exactly as you documented. The direction here is right, all three hunks are load-bearing, and I reproduced your claimed 6-of-9 revert-failure exactly — the three survivors really are the controls you said they were.

Adversarial review found two things that need a fix pass before merge.

**Blocker 1 — merged alone, this converts `#N/A` into a silent `0`.**

Your new temporal arm in `value_to_f64_lenient` sits directly above the untouched `LiteralValue::Empty => Some(0.0)` arm. With `A1:A5` date-typed and `A6:A10` never written:

| formula | base `fe7b9111` | **#321 alone** | #322 alone | #321+#322 | Excel |
|---|---|---|---|---|---|
| `=VLOOKUP(D1,A1:B10,2,TRUE)` | `#N/A` | **`0`** | `#N/A` | `30` | `30` |

Before your change `cmp_for_lookup(Date, Date)` was `None`, the binary search collapsed, and you got `#N/A`. After it the dates are comparable but the trailing blanks still read as serial `0`, so the vector is `45292..45296, 0,0,0,0,0`, the bisection walks onto a blank row, and egress publishes it as numeric `0` per #279. That is issue #317's failure mode, newly extended to date columns — so the PR body's "the change only converts `#N/A` into the value Excel produces" is not quite true.

It is fine merged **together with #322**, which I verified. So this is an ordering constraint rather than a defect in your logic: the two need to land as a pair, and it is worth stating that explicitly.

**Blocker 2 — the 1900 date system is hardcoded in a crate that is otherwise date-system aware.**

All three sites use `as_serial_number()`, which is `as_serial_number_for(DateSystem::Excel1900)`. Elsewhere the crate threads the workbook's system — `builtins/datetime/date_time.rs:112`, `interpreter.rs:1177`, `formula_plane/span_eval.rs:1118` — and there is a test named `temporal_overlay_conversion_uses_the_workbook_date_system`. 1904 is reachable in production through the JSON backend (`backends/json.rs:677`) and the Python `date_system="1904"` API (`bindings/python/src/engine.rs:275`).

Under a 1904 workbook the fix half-lands, and worse, introduces a new false positive:

| formula | base | **#321** | Excel (1904) |
|---|---|---|---|
| `=MATCH(D1,A1:A5,1)` | `#N/A` | `3` | `3` |
| `=MATCH(D1,A1:A5,0)` | `#N/A` | **`#N/A`** | `3` |
| `=MATCH(D1,F1:F1,0)` where `F1 = 45294` | `#N/A` | **`1`** | `#N/A` |

The cells reach the exact path through the Arrow numeric lane holding 1904 serials, while the needle converts through the hardcoded 1900 helper. Approximate mode works by accident because both sides use the same wrong helper. `45294` under 1904 is 2028-01-04, so a 2024 date matches a 2028 serial.

`value_to_f64_lenient` is a free function with no context, so this is a design question rather than a one-liner. I would rather thread `DateSystem` through the lookup helpers than scope the PR to 1900 — an unfixed case is acceptable, a new wrong answer is not.

**Major — hunk 3 is pinned by zero tests.** `LookupHashKey::from_literal` is demonstrably live, and when it is absent the same formula can return different answers depending on whether the hash index is warm. That is the most alarming failure mode in the PR and it has no coverage. Mutant N11 (revert hunk 3) survives your suite and the full 2611-test suite.

**Minor worth knowing.** `cmp_for_lookup`'s absolute tolerance is `1e-12`, but one ULP at date-serial magnitudes is about `7.3e-12` — so at these magnitudes the tolerance is narrower than a ULP and effectively inert. It works today because both sides route through the same helper, but it means temporal equality is bit-exact by accident rather than by design.

Mutation testing: 15 mutants against your three hunks, of which N02 (drop `DateTime`/`Time`), N05/N08/N12 (`serial.trunc()`), N11 (revert hunk 3), N14 (use 1904), and N15 (`Empty => None`) survive with confirmed distinguishing inputs.

None of this is a reason to reject. It is careful, honestly documented, correctly oracled work, and the two blockers are exactly the kind that only surface under attack. Happy for you to take the fix pass if you would like it — say the word — otherwise I will pick it up so this can ship in 0.8.2 alongside #322 and #320.

PSU3D0:
Update so you are not left waiting or duplicating effort: I am taking the fix pass on both PRs now, so this can ship in 0.8.2 today alongside #318, which is already merged.

Both PRs have "allow edits by maintainers" enabled, so I am pushing the fixes as additional commits onto your own branches rather than opening replacement PRs. Your commits stay exactly as authored — I am not rebasing, squashing, or amending anything — and both PRs merge under your name.

Two design calls were mine to make and are now settled, so you know what is landing on top of your work:

**Thread `DateSystem` through the lookup helpers** rather than scoping to the 1900 system. Your three sites are correct in what they do, they just need the workbook's date system instead of the hardcoded one, which means threading a parameter through the call sites of a free function. An unfixed 1904 case would have been acceptable; the new false positive it created was not.

**Errors propagate** from an approximate-lookup range rather than being skipped. Your comparability test made them unselectable as a side effect, which is a stronger rule than the module doc claims, so we are making it explicit in the other direction and documenting it.

I am also folding #320 into #322's branch, since the review confirmed the one-line change belongs in the function you are already editing, and it interacts with your sortedness-guard repair.

Merge order will be #322 first, then #321, because #322 alone is safe while #321 alone regresses `VLOOKUP` over a date column with a blank tail to `0`. A test pinning that pairing is going onto #321's branch so the constraint cannot be lost later.

Genuinely good work on both — five real defects found, correctly oracled, honestly scoped, with the ones you deliberately did not fix filed separately as #319 and #320 rather than smuggled in. #319 is deferred on purpose: `Empty -> 0.0` is load bearing across #279 and #285, so it wants an Excel and LibreOffice variant study across exact match, approximate match, criteria, and aggregation before anyone touches that helper. If that interests you it is a good next one, and I would rather agree the semantics table with you up front than review it afterwards.

PSU3D0:
Landed in #325, which carries your commit `f4d816b4` with your authorship intact. Closing this PR as merged-via-#325 rather than rejected — the change is on `main` and shipping in 0.8.2.

Two mechanical reasons it could not merge here. It conflicted with #322 (both edit `lookup_utils.rs`), and the fix pass could not be pushed onto your branch: GitHub does not honour `maintainer_can_modify` for **organization-owned** forks, and `Go-Digital-Alchemy-Repos` is an organisation. The API cheerfully reports `maintainer_can_modify=true` anyway, which is how I found out the hard way. Worth knowing for future contributions — a user-owned fork would have let me push the fix pass straight onto your branch and merge this PR normally.

Your #322 and #323 merged directly and unmodified.

What landed on top of your work in #325: the workbook date system threaded through the lookup helpers rather than the hardcoded 1900 (your three sites were right about what to do, they just needed the workbook's system — under 1904 the original made a 2024 date match a 2028 serial); tests for `LookupHashKey::from_literal`, which had none; a test pinning the #321-plus-#322 pairing so the blank-tail-to-zero hazard cannot come back; and mutant-killing tests for the boundaries the review found unpinned.

Net for 0.8.2 from your five reports: #316, #317, #318 and #320 all fixed, #319 deferred deliberately. That is a genuinely unusual first contribution — thank you.

Issue #283, opened 2026-08-06
[bug] eval: VLOOKUP/HLOOKUP approximate mode has no sortedness guard; behavior contradicts module doc and MATCH


PSU3D0:
The VLOOKUP/HLOOKUP approximate-match arms call the binary search with no sortedness guard, while MATCH implements ascending-sort detection and the lookup module doc claims unsorted data leads to `#N/A` like Excel. On descending data the engine silently returns wrong-row values: `=VLOOKUP(2.5,{3,"c";2,"b";1,"a"},2,TRUE)` returns `"a"` while LibreOffice returns `#N/A`.

This is pre-existing (found during the #280 adversarial review), but #280's Excel-aligned default flip routes every three-argument VLOOKUP/HLOOKUP call into this unguarded path, so the exposure is now materially higher.

There is a genuine behavioral decision to make before fixing: Excel itself does not guard — its documentation says approximate results on unsorted data "may not be correct", i.e. Excel returns garbage rather than `#N/A`, whereas LibreOffice detects and errors, and our MATCH guards like LibreOffice. Options: add the `is_sorted` guard to the V/HLOOKUP approximate arms (LO-consistent, safer, internally consistent with MATCH and the module doc), or match Excel's unguarded behavior exactly and fix the module doc to stop claiming `#N/A`. Either way the current state — doc says one thing, MATCH does another, V/HLOOKUP do a third — should not survive into 0.8.0 unexamined. Real-Excel verification of the descending-data case is warranted before choosing, since LO is a known-imperfect oracle for guard behavior.

The docs on the #280 branch already warn that approximate matching assumes ascending-sorted data, so this issue is about the runtime behavior only.

drafted with love by (agent) Manfred, with approval from Frankie

PSU3D0:
Decision: approximate-mode lookups guard on sortedness and return `#N/A` on unsorted data, matching LibreOffice and our existing MATCH behaviour. Excel documents its result on unsorted data as possibly incorrect, so there is no defined result to be parity with, and for pipeline consumers a loud `#N/A` beats a silent wrong row. The cost objection does not hold: the approximate arm already clones the column per call (#414), so the guard measured 1.02 to 1.04 times at 200k rows.

[...]

Issue #318, opened 2026-08-11
[bug] python: every published wheel has TODAY()/NOW() frozen at 1970-01-01


tommy230:
### Summary

`bindings/python/Cargo.toml` depends on `formualizer` with
`default-features = false` and a feature list that omits `system-clock`.
`FunctionContext::clock` then falls back to the `FixedClock` at the UTC
epoch, so every volatile date/time builtin answers 1970-01-01 in every
published wheel.

### Reproduction

```python
import formualizer as fz

wb = fz.Workbook(mode=fz.WorkbookMode.Ephemeral)
wb.add_sheet("Sheet1")
wb.set_formula("Sheet1", 1, 1, "=TODAY()")
wb.set_formula("Sheet1", 2, 1, "=YEAR(TODAY())")
wb.set_formula("Sheet1", 3, 1, "=NOW()")
print(wb.evaluate_cell("Sheet1", 1, 1))
print(wb.evaluate_cell("Sheet1", 2, 1))
print(wb.evaluate_cell("Sheet1", 3, 1))
```

### Expected behavior

`TODAY()` returns today's serial; `NOW()` returns today's serial plus a time
fraction. The Rust crate's own default feature set includes `system-clock`,
so the binding is the odd one out.

### Actual behavior

Measured on a wheel built from `fe7b9111`, macOS arm64 abi3, on 2026-08-11:

```
TODAY()        -> 25569.0            # 1899-12-30 + 25569 days = 1970-01-01
YEAR(TODAY())  -> 1970.0
NOW()          -> 25569.0            # no time fraction at all
```

### Additional context

Silently wrong rather than an error, so nothing downstream notices: any
`=IF(TODAY()>...)`, ageing bucket, or accrual-to-date in a customer workbook
evaluates against 1970. One-line fix plus regression tests ready — will open
a PR against this issue. The emscripten/Pyodide target should stay as it is;
ambient wall-clock time is not part of the portable wasm profile.

### Environment

- Formualizer version: 0.8.1 (`main` at `fe7b9111`)
- Host: Python
- OS: macOS 15 (Darwin 25.5.0), aarch64
- Compiler / Python: rustc 1.93.0 (CI-pinned toolchain) / CPython 3.12.13
- Enabled features: `--all-features`

Oracle for every expected value: Microsoft Excel 16.105.3 (Microsoft 365 for Mac), measured on throwaway synthetic workbooks.

Issue #310, opened 2026-08-10
[bug]  Date − Number returns #NUM! when the serial result is negative (e.g. =DATE(2024,12,1)-45658)


Devider:
## Summary

In Excel/LibreOffice a date is stored and computed as a serial number, so subtracting a larger number from a date simply yields a negative number (e.g. `2024-12-01` is serial `45627`, and `45627 − 45658 = −31`).

formualizer `0.8.0` represents dates as a distinct value type and returns `#NUM!` whenever a `Date − Number` subtraction would produce a **negative** serial, instead of falling back to a plain Number. All other arithmetic operators (`+`, `*`, `/`, `^`) coerce the date to a serial and work; only binary `-` with a negative result fails.

This breaks the common financial idiom `(end_date − start_date)/365 * ...` whenever `end_date < start_date` (e.g. a payment date before the start-of-year anchor): a perfectly fine negative fraction-of-year becomes `#NUM!` and propagates through every downstream cell.

## Environment

- formualizer `0.8.0` (PyPI, released 2026-08-10)
- Python 3.x, Linux
- Reproduced via `load_workbook` + `evaluate_cell`; cross-checked against LibreOffice 24.2 headless UNO recalculation

## Reproduction

Minimal workbook `negdate.xlsx` — `A1 = DATE(2024,12,1)` (serial 45627), `B1 = 45658`:

| Cell | Formula | LibreOffice / Excel | formualizer 0.8.0 |
|---|---|---|---|
| A1 | `DATE(2024,12,1)` | `45627` | `45627` |
| B1 | `45658` | `45658` | `45658` |
| C1 | `=A1-B1` | **`-31`** | **`#NUM!`** ❌ |
| D1 | `=B1-A1` | `31` | `31` ✅ |
| E1 | `=A1+B1` | `91285` | `91285` ✅ |
| F1 | `=A1*2` | `91254` | `91254` ✅ |
| G1 | `=MAX(IF(YEAR(A1)>2025,A1*0,(A1-B1)/365*0),0)` | **`0`** | **`#NUM!`** ❌ |

Direct engine probes on a date cell `F132 = 2024-12-01` and `S1 = 45658`:

```
=F132 - S1      -> #NUM!        (=F132-45658 also #NUM!)
=F132 - 45627   -> 1899-12-31   (serial 0 is representable as a Date)
=S1 - F132      -> 31.0
=F132 + S1      -> 2149-12-04   (serial 91285, boxed as Date)
[...]

Issue #280, opened 2026-08-06
[bug] eval: VLOOKUP/HLOOKUP range_lookup mishandles numeric 0 and omitted slots; absent default diverges from Excel


PSU3D0:
VLOOKUP's `range_lookup` argument diverges from Excel in three distinct ways at `95b67049`, verified against LibreOffice 24.2.7 headless recalc today with sorted data `D1:E3` = `{1,"a";2,"b";3,"c"}` and lookup value `2.5`.

First, a numeric `range_lookup` of `0` is treated as approximate: `=VLOOKUP(2.5,D1:E3,2,0)` returns `"b"` in Formualizer but `#N/A` in LibreOffice and Excel, where numeric zero means exact match. The implementation matches only `LiteralValue::Boolean` and falls through to approximate for every other value type.

Second, an explicitly omitted slot is treated as approximate: `=VLOOKUP(2.5,D1:E3,2,)` returns `"b"` but should be exact (`#N/A`). After #277, the omitted slot materializes as numeric zero and hits the same fall-through arm, so this is the same coercion defect surfacing through omission.

Third, the absent-argument default is an intentional, documented divergence: Formualizer defaults to exact (`FALSE`) for safety, while Excel and LibreOffice default to approximate (`TRUE`): `=VLOOKUP(2.5,D1:E3,2)` returns `#N/A` here but `"b"` there. Ahead of an 0.8.0 whose contract is a documented Excel-compatible subset, a silent behavioral divergence on one of the most-used functions works against the product promise; this issue proposes aligning the default to Excel's approximate behavior and documenting the change prominently, since existing users may rely on the safer default.

HLOOKUP shares the same argument handling and needs the same fixes. `=VLOOKUP(2.5,D1:E3,2,TRUE)`, `=...,FALSE)`, and `=...,1)` already match the oracle.

drafted with love by (agent) Manfred, with approval from Frankie

PR #322, opened 2026-08-11
fix(eval): approximate lookup ignores blanks and other-class entries (#317)

### LINKED ISSUE #317: [bug] eval: approximate lookup treats blank cells as zero, so an over-wide range returns #N/A (or a blank row)
### Summary

Excel's legacy approximate lookups ignore entries outside the needle's value
class — blank cells, and a text header above a column of numbers. Formualizer
coerces blanks to numeric `0` and compares every entry, so a lookup range
that extends past its populated rows looks unsorted and `MATCH` returns
`#N/A`, while `VLOOKUP` (which has no sortedness guard) selects a blank row
and publishes `0`.

Pointing a lookup at a whole column, or at a range sized for future rows, is
about as common as spreadsheet idioms get, so the blast radius is wide.

### Reproduction

2. `B1:B5` = `10, 20, 30, 40, 50`.
3. `C1` = the text `"Header"`; `C2:C6` = `1, 2, 3, 4, 5`.
4. `G1,G2,G4,G5` = `1, 2, 4, 5`; leave `G3` blank.
5. `I1:I4` = `"apple", 1, 2, "zebra"`.
6. Evaluate the formulas below.

### Minimal input

| --- | --- | --- |
| `=MATCH(3,A1:A10,1)` | 3 | `#N/A` |
| `=MATCH(3.5,A1:A10,1)` | 3 | `#N/A` |
| `=MATCH(6,A1:A10,1)` | 5 | `#N/A` |
| `=VLOOKUP(3,A1:B10,2,TRUE)` | 30 | `0` |
| `=MATCH(4,G1:G5,1)` | 4 | `#N/A` |
| `=MATCH(3,G1:G5,1)` | 2 | `#N/A` |
| `=MATCH(3,C1:C6,1)` | 4 | `#N/A` |
| `=MATCH(2,I1:I4,1)` | 3 | `#N/A` |
| `=MATCH("m",I1:I4,1)` | 1 | `#N/A` |

Note `=MATCH(3,C1:C6,1)` => **4**, not 3: the skipped header does not shift
the answer, because Excel counts the position in the original range.

### Expected behavior

The approximate search visits only the entries in the needle's value class.
Skipped entries are neither out-of-order data nor matchable positions, and
the returned index is a position in the original range.

### Actual behavior

`value_to_f64_lenient` maps `Empty` to `0.0`, so `1,2,3,4,5,blank,blank,...`
reads as `1,2,3,4,5,0,0,...`. `MATCH`'s `is_sorted_ascending` check rejects
it; `VLOOKUP`/`HLOOKUP` have no such check and bisect the misordered vector,
landing on a blank row that egress then publishes as numeric `0`.

### Additional context

This is *not* the missing-sortedness-guard question in #283, and a fix for it
does not add a guard to `VLOOKUP`/`HLOOKUP`; #283 stays open either way. Fix
and regression tests ready — will open a PR against this issue.

### Environment

- Host: Rust

tommy230:
Fixes #317.

## What was wrong

Point an approximate lookup at a range wider than its populated rows — a
whole column, or a range sized for rows that do not exist yet — and it
breaks:

- `MATCH(3, A1:A10, 1)` over `A1:A5 = 1..5` with `A6:A10` blank returns
  `#N/A`. `value_to_f64_lenient` maps `Empty` to `0.0`, so `is_sorted_ascending`
  sees `1,2,3,4,5,0,0,0,0,0`, calls it unsorted, and bails.
- `VLOOKUP(3, A1:B10, 2, TRUE)` over the same data returns `0`. There is no
  sortedness guard on that path (deliberately — see #283), so the binary
  search runs over the misordered vector, lands on a blank row, and egress
  publishes the blank as numeric `0`. A wrong number, silently.

The same coercion breaks the type-mismatch case. A text header above a
numeric column, or a stray number inside a text column, is compared rather
than skipped, and `MATCH` calls the range unsorted.

## Excel behaviour

Excel's legacy approximate lookups (`MATCH` with `match_type` 1 or -1,
`VLOOKUP`/`HLOOKUP` with `range_lookup` TRUE) search only the entries in the
needle's value class. Blanks are skipped. Entries of another class are
skipped. Neither counts as out-of-order data, and neither is a matchable
position. Crucially, the returned index is still a position in the
**original** range, not in the compacted one.

Measured on Microsoft Excel 16.105.3 (Microsoft 365 for Mac), synthetic
workbook. `A1:A5` = 1..5 with `A6:A10` blank; `B1:B5` = 10..50; `C1` =
`"Header"` with `C2:C6` = 1..5; `E1:E5` = 5..1 descending with `E6:E10`
blank; `G` = `1, 2, blank, 4, 5`; `I` = `"apple", 1, 2, "zebra"`; `K1:K3` =
`"alpha","beta","gamma"` with a blank tail; `M` = `3,1,5,2,4`:

| Formula | Excel | main @ `fe7b9111` |
| --- | --- | --- |
| `=MATCH(3,A1:A10,1)` | 3 | `#N/A` |
| `=MATCH(3.5,A1:A10,1)` | 3 | `#N/A` |
| `=MATCH(6,A1:A10,1)` | 5 | `#N/A` |
| `=VLOOKUP(3,A1:B10,2,TRUE)` | 30 | `0` |
| `=MATCH(4,G1:G5,1)` | 4 | `#N/A` |
| `=MATCH(3,G1:G5,1)` | 2 | `#N/A` |
| `=MATCH(3,C1:C6,1)` | 4 | `#N/A` |
| `=MATCH(3.5,C1:C6,1)` | 4 | `#N/A` |
| `=MATCH(2,I1:I4,1)` | 3 | `#N/A` |
| `=MATCH("m",I1:I4,1)` | 1 | `#N/A` |
| `=MATCH("beta",K1:K10,1)` | 2 | `#N/A` |
| `=MATCH(2,M1:M5,1)` (control) | `#N/A` | `#N/A` |
| `=MATCH(0.5,A1:A10,1)` (control) | `#N/A` | `#N/A` |

Two rows are worth dwelling on. `=MATCH(3,C1:C6,1)` is **4**, not 3: the
skipped header does not shift the answer, which is the direct evidence that
Excel maps back to the original position. And `=MATCH("m",I1:I4,1)` is 1,
the mirror direction — a text needle skips the numbers.

The unsorted control matters too. Excel does not silently guess on genuinely
unsorted data here; it returns `#N/A`, which is what `MATCH` already does and
what this PR keeps doing.

## The fix

`SearchedVector` (in `lookup_utils.rs`) projects a lookup vector onto the
entries an approximate search visits, and maps an index in that projection
back to its row in the original range.

The projection is lazy about allocating: it scans for the first entry that
would be skipped and, if there is none, borrows the slice and stores no
positions at all. The common case — a vector entirely in the needle's class —
costs one pass and zero allocation, which matters because these paths already
materialize the vector and are on the hot lookup route.

`binary_search_match` builds a `SearchedVector` internally and returns an
original-range index, so all six approximate call sites are covered at once:
`MATCH`, and both the reference and the array-literal paths of `VLOOKUP` and
`HLOOKUP`. `MatchFn` additionally runs its sortedness guard over the same
projection, so the guard and the search can no longer disagree about what is
being searched, and its small-vector linear scan iterates the projection too.

What does **not** change:

- No sortedness guard is added to `VLOOKUP`/`HLOOKUP`. #283 is a separate
  question and stays open; the module doc's account of the deliberate
  difference between `MATCH` and `VLOOKUP` is still accurate.
- Exact match is untouched. `find_exact_index` and friends see the same
  values they always did.
- The tie-breaking and direction logic inside the search is unchanged. Only
  *which entries* it visits changed.

## Scope

Not a regression; the coercion is long-standing. The blast radius is wide
because the trigger is an ordinary spreadsheet habit: sizing a lookup range
past the current data, or keeping a header row inside the range. Every such
formula either errors or, on the `VLOOKUP` path, quietly returns zero.

## Verification

Fourteen regression tests in
`crates/formualizer-eval/src/engine/tests/approximate_lookup_ignored_entries.rs`,
each pinned to a measured Excel result. Revert the production change and
**ten of the fourteen fail**. The four survivors are controls:

- genuinely unsorted data is still `#N/A` (the fix must not weaken the guard),
- a needle below every key is still `#N/A`,
- exact match over the same blank-tailed range is unchanged,
- the descending blank-tail case, which `main` already answers correctly —
  but only because the needle hits a key exactly and the linear scan returns
  early, so it is a control rather than evidence.

Gates, run locally on the CI-pinned toolchain (macOS 15, aarch64):

```
cargo +1.93.0 fmt --all -- --check                                  exit 0
cargo +1.93.0 clippy --workspace --exclude formualizer-bench-core \
  --exclude formualizer-python --exclude formualizer-wasm \
  --all-targets --all-features -- -D warnings                       exit 0
cargo +1.93.0 test --workspace --exclude formualizer-bench-core \
  --exclude formualizer-python --exclude formualizer-wasm \
  --all-features --no-fail-fast            3476 passed, 1 failed
```

The single failure is pre-existing and unrelated: the JSON formula suite's
`IMEXP`/`IMSIN`/`IMCOS` cases differ in the last ULP on macOS arm64.
Unmodified `main` gives 3462 passed, 1 failed on the same host.

Not run here: the Linux-only `public-api` snapshot job. `SearchedVector` and
`is_searchable_for_approximate` are added to
`builtins::lookup::lookup_utils`, which does not appear in
`public-api/formualizer-eval.txt`; no exported item changes.

- **`MATCH` with `match_type = -1` on 8+ descending entries** (#320):
  `binary_search_match`'s `-1` branch selects the *first* entry `>=` the
  needle where descending data wants the *last*, so
  `=MATCH(3.5,A1:A10,-1)` over `10..1` returns 1 where Excel returns 7. The
  `< 8` linear path gets it right, and an exact hit masks it. Verified
  present on unmodified `main`; this PR changes which entries that branch
  visits, not how it chooses among them, so the defect is orthogonal and
  gets its own fix.
- **Exact `MATCH` coercing blanks to zero** (#319): `=MATCH(0,A1:A10,0)`
  returns 6 where Excel returns `#N/A`. Different code path; and
  `Empty -> 0.0` is load-bearing elsewhere (#279, #285), so narrowing it
  deserves its own decision.
- **A sortedness guard for `VLOOKUP`/`HLOOKUP`** (#283). Unchanged here.
- **`XLOOKUP`/`XMATCH`.** `approximate_select_ascending` and
  `guard_sorted_ascending` still see blanks as zero. Those functions have
  their own match-mode semantics and deserve their own oracle pass rather
  than being swept in.

Drafted by Claude (agent), verified by execution against a real Excel
oracle, reviewed by a human before submission.

### DIFF
--- a/crates/formualizer-eval/src/builtins/lookup/core.rs
+++ b/crates/formualizer-eval/src/builtins/lookup/core.rs
@@ -15,7 +15,7 @@
 //! - Type coercion: current simple: numbers vs numeric text coerced; text comparison case-insensitive? Excel is case-insensitive for MATCH (without wildcards). We implement case-insensitive for now.
 //!   TODO(excel-nuance): refine boolean/text/number coercion differences.
 
-use super::lookup_utils::{cmp_for_lookup, find_exact_index, is_sorted_ascending};
+use super::lookup_utils::{SearchedVector, cmp_for_lookup, find_exact_index};
 use crate::args::{ArgSchema, CoercionPolicy, ShapeKind};
 use crate::engine::lookup_index_cache::LookupAxis;
 use crate::function::Function;
@@ -24,18 +24,38 @@ use formualizer_common::ArgKind;
 use formualizer_common::{ExcelError, ExcelErrorKind, LiteralValue};
 use formualizer_macros::func_caps;
 
+/// Approximate search over a lookup vector, returning a position in the
+/// *original* vector.
+///
+/// Entries the search must ignore — blanks, and entries outside the needle's
+/// value class — are projected out first, so they neither occupy a matchable
+/// position nor disturb the binary search's ordering assumption.
 fn binary_search_match(slice: &[LiteralValue], needle: &LiteralValue, mode: i32) -> Option<usize> {
     if mode == 0 || slice.is_empty() {
+        return None;
+    }
+    let searched = SearchedVector::new(slice, needle);
+    binary_search_searched(&searched, needle, mode).map(|i| searched.original_position(i))
+}
+
+/// Same search, but over an already-projected vector and returning an index
+/// into that projection.
+fn binary_search_searched(
+    searched: &SearchedVector<'_>,
+    needle: &LiteralValue,
+    mode: i32,
+) -> Option<usize> {
+    if mode == 0 || searched.is_empty() {
         return None;
     }
     // Only ascending binary search currently (mode 1); descending path kept linear for now.
     if mode == 1 {
         // largest <= needle
         let mut lo = 0usize;
-        let mut hi = slice.len();
+        let mut hi = searched.len();
         while lo < hi {
             let mid = (lo + hi) / 2;
-            match cmp_for_lookup(&slice[mid], needle) {
+            match cmp_for_lookup(searched.get(mid), needle) {
                 Some(c) => {
                     if c > 0 {
                         hi = mid;
@@ -52,8 +72,8 @@ fn binary_search_match(slice: &[LiteralValue], needle: &LiteralValue, mode: i32)
     } else {
         // -1 mode handled via linear fallback since semantics differ (smallest >=)
         let mut best: Option<usize> = None;
-        for (i, v) in slice.iter().enumerate() {
-            if let Some(c) = cmp_for_lookup(v, needle) {
+        for i in 0..searched.len() {
+            if let Some(c) = cmp_for_lookup(searched.get(i), needle) {
                 if c == 0 {
                     return Some(i);
                 }
@@ -267,13 +287,16 @@ impl Function for MatchFn {
                         return Ok(crate::traits::CalcValue::Scalar(LiteralValue::Error(e)));
                     }
 
+                    // Project out the entries an approximate search ignores
+                    // (blanks and entries outside the needle's value class)
+                    // before both the sortedness guard and the search itself.
+                    let searched = SearchedVector::new(&values, &lookup_value);
+
                     // Lightweight unsorted detection for approximate modes
                     let is_sorted = if mt == 1 {
-                        is_sorted_ascending(&values)
+                        searched.is_sorted_ascending()
                     } else if mt == -1 {
-                        values
-                            .windows(2)
-                            .all(|w| cmp_for_lookup(&w[0], &w[1]).is_some_and(|c| c >= 0))
+                        searched.is_sorted_descending()
                     } else {
                         true
                     };
@@ -282,32 +305,31 @@ impl Function for MatchFn {
                             ExcelError::new(ExcelErrorKind::Na),
                         )));
                     }
-                    let idx = if values.len() < 8 {
+                    let idx = if searched.len() < 8 {
                         // linear small
-                        let mut best: Option<(usize, &LiteralValue)> = None;
-                        for (i, v) in values.iter().enumerate() {
-                            if let Some(c) = cmp_for_lookup(v, &lookup_value) {
+                        let mut best: Option<usize> = None;
+                        for i in 0..searched.len() {
+                            if let Some(c) = cmp_for_lookup(searched.get(i), &lookup_value) {
                                 // compare candidate to needle
                                 if mt == 1 {
                                     // v <= needle
-                                    if (c == 0 || c == -1)
-                                        && (best.is_none() || i > best.unwrap().0)
+                                    if (c == 0 || c == -1) && (best.is_none() || i > best.unwrap())
                                     {
-                                        best = Some((i, v));
+                                        best = Some(i);
                                     }
                                 } else {
                                     // -1, v >= needle
-                                    if (c == 0 || c == 1) && (best.is_none() || i > best.unwrap().0)
-                                    {
-                                        best = Some((i, v));
+                                    if (c == 0 || c == 1) && (best.is_none() || i > best.unwrap()) {
+                                        best = Some(i);
                                     }
                                 }
                             }
                         }
-                        best.map(|(i, _)| i)
+                        best
                     } else {
-                        binary_search_match(&values, &lookup_value, mt)
+                        binary_search_searched(&searched, &lookup_value, mt)
                     };
+                    let idx = idx.map(|i| searched.original_position(i));
                     match idx {
                         Some(i) => Ok(crate::traits::CalcValue::Scalar(LiteralValue::Int(
                             (i + 1) as i64,
--- a/crates/formualizer-eval/src/builtins/lookup/lookup_utils.rs
+++ b/crates/formualizer-eval/src/builtins/lookup/lookup_utils.rs
@@ -113,6 +113,79 @@ pub fn equals_maybe_wildcard(
     PreparedLookupMatcher::new(pattern, wildcard).matches(candidate)
 }
 
+/// Whether Excel's approximate search visits `value` when looking for `needle`.
+///
+/// The legacy approximate lookups (`MATCH` with `match_type` 1/-1,
+/// `VLOOKUP`/`HLOOKUP` with `range_lookup` TRUE) consider only entries in the
+/// needle's value class. A blank cell, or an entry of another class such as a
+/// text header sitting above a column of numbers, is skipped: it is neither
+/// out-of-order data nor a matchable position.
+pub fn is_searchable_for_approximate(value: &LiteralValue, needle: &LiteralValue) -> bool {
+    !matches!(value, LiteralValue::Empty) && cmp_for_lookup(value, needle).is_some()
+}
+
+/// A lookup vector projected onto the entries an approximate search visits.
+///
+/// Positions are only materialized when something is actually skipped, so the
+/// common case — a vector that is entirely in the needle's class — borrows the
+/// original slice and allocates nothing. Indices returned by a search over this
+/// projection are mapped back with [`SearchedVector::original_position`],
+/// because Excel counts the answer from the top of the *original* range.
+pub struct SearchedVector<'a> {
+    values: &'a [LiteralValue],
+    positions: Option<Vec<usize>>,
+}
+
+impl<'a> SearchedVector<'a> {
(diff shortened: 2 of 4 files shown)


PSU3D0:
This is the stronger of your two PRs, and the `SearchedVector` design is sound. I independently confirmed the subtlest property you claimed — that the returned index is a position in the **original** range, not the compacted one — through both the linear and the binary path, with leading skips, interior skips, sparse 50%-blank vectors, and duplicate keys. It also fixes materially more than the PR claims: HLOOKUP blank tails, the #280 default-`range_lookup` path, all-blank ranges, and descending blank tails.

Two things block merge.

**Blocker — approximate lookups now silently swallow error cells.**

`is_searchable_for_approximate` is `!matches!(value, Empty) && cmp_for_lookup(value, needle).is_some()`, and `cmp_for_lookup(Error(..), _)` is `None`. So an error entry is projected out exactly like a text header: it cannot break sortedness and can never be returned.

With `H1=1, H2=2, H3=#DIV/0!, H4=9`:

| formula | base | **#322** | Excel |
|---|---|---|---|
| `=MATCH(5,H1:H4,1)` | `#N/A` | **`2`** | an error, not a silent number |

This is reachable without anyone writing an error on purpose: pre-1900 date cells materialise as `#NUM!` in this engine, which #310/#311 made newly consequential by making negative serials reachable. With `A1:A4` = dates 1899-01-01, 1899-06-01, 1900-03-01, 2024-01-01, `=MATCH(50000,A1:A4,1)` returns `4` on #321+#322 where every branch previously returned `#N/A`. The answer is numerically what Excel would give if the cells were valid — but it is produced by discarding two `#NUM!` cells without a word.

The module doc at `core.rs:19` currently says errors in non-deciding positions "don't matter unless selected". Your change makes them never selectable, which is a strictly stronger and undocumented rule. Whichever way we go, it needs to be a stated decision with a test, not a side effect of the comparability check. My inclination is that errors should propagate rather than be skipped, but I want an oracle on it before asking you to implement either.

**Major — this widens #320's blast radius, and "orthogonal" understates it.**

The projection changes both whether the descending sortedness guard bails and whether `searched.len() < 8` selects `MatchFn`'s correct linear path or the defective binary one. The effect runs in both directions:

| fixture | formula | base | **#322** | Excel |
|---|---|---|---|---|
| `E1:E5 = 5..1`, `E6:E10` blank | `=MATCH(3.5,E1:E10,-1)` | `1` | **`2`** | `2` |
| `G1:G11 = 9,8,7,_,6,5,4,_,3,2,1` | `=MATCH(5.5,G1:G11,-1)` | `#N/A` | **`1`** | `5` |

On `main` the interleaved blanks broke the descending guard and the formula errored. You repair the guard — correctly — the search then runs, and #320's "first `>=`" defect produces a wrong number where a wrong number was previously impossible. Not a reason to reject, since the guard repair is right and the underlying defect is #320's, but the PR body should say it: this moves #320 from latent on blank-free descending ranges to live on any descending range with 8 or more searchable entries.

**The good news is that #320 becomes a one-liner on top of this.** Flipping `best.is_none_or(|b| i < b)` to `i > b` at `core.rs:78` makes all three of my #320 probes return the Excel answer (`1 -> 7`, `1 -> 7`, `1 -> 5`) and breaks nothing in the 2616-test suite. Since you are already in this function, folding #320 in here is cleaner than a separate PR.

**Minors.** The `<8` threshold now means 8 *searchable* elements, which is behaviourally significant and undocumented (`core.rs:11`). The error-propagation sentence at `core.rs:19` is now wrong. `binary_search_searched`'s `None => hi = mid` arm is unreachable by construction, silently encoding an invariant instead of asserting it. And your "value class" model is a comparability test rather than a class test, so booleans and numeric-looking text are not separated from numbers — weaker than the doc comment implies.

Mutation testing: 23 mutants, 16 killed, 7 survived, 2 of those provably equivalent. Live survivors worth pinning are M10 (ascending guard `c <= 0` to `c < 0`), M14 (drop the `original_position` remap), M15 (`searched.len()` to `values.len()` at the threshold), and M18 (the #320 fix itself).

Genuinely good work — the original-position property is the hard part and you got it right on both paths. Happy for you to take the fix pass if you want it; otherwise I will, so this can ship in 0.8.2 with #321 and #320.

PSU3D0:
Update so you are not left waiting or duplicating effort: I am taking the fix pass on both PRs now, so this can ship in 0.8.2 today alongside #318, which is already merged.

Both PRs have "allow edits by maintainers" enabled, so I am pushing the fixes as additional commits onto your own branches rather than opening replacement PRs. Your commits stay exactly as authored — I am not rebasing, squashing, or amending anything — and both PRs merge under your name.

Two design calls were mine to make and are now settled, so you know what is landing on top of your work:

**Thread `DateSystem` through the lookup helpers** rather than scoping to the 1900 system. Your three sites are correct in what they do, they just need the workbook's date system instead of the hardcoded one, which means threading a parameter through the call sites of a free function. An unfixed 1904 case would have been acceptable; the new false positive it created was not.

**Errors propagate** from an approximate-lookup range rather than being skipped. Your comparability test made them unselectable as a side effect, which is a stronger rule than the module doc claims, so we are making it explicit in the other direction and documenting it.

I am also folding #320 into #322's branch, since the review confirmed the one-line change belongs in the function you are already editing, and it interacts with your sortedness-guard repair.

Merge order will be #322 first, then #321, because #322 alone is safe while #321 alone regresses `VLOOKUP` over a date column with a blank tail to `0`. A test pinning that pairing is going onto #321's branch so the constraint cannot be lost later.

Genuinely good work on both — five real defects found, correctly oracled, honestly scoped, with the ones you deliberately did not fix filed separately as #319 and #320 rather than smuggled in. #319 is deferred on purpose: `Empty -> 0.0` is load bearing across #279 and #285, so it wants an Excel and LibreOffice variant study across exact match, approximate match, criteria, and aggregation before anyone touches that helper. If that interests you it is a good next one, and I would rather agree the semantics table with you up front than review it afterwards.

Issue #239, opened 2026-07-26
CELL and INFO are unimplemented (and need a determinism decision first)


PSU3D0:
`CELL` and `INFO` return `#NAME?` on `main` (9b8fb52a).

| formula | current |
|---|---|
| `=CELL("row",A1)` | `#NAME?` |
| `=INFO("numfile")` | `#NAME?` |

These are the last two commonly-expected information functions that remain unimplemented. The other outstanding cluster is the array/lambda family, filed separately.

## Worth deciding before implementing

Both are environment-introspection functions, which sits awkwardly with reproducibility. `INFO("directory")`, `INFO("osversion")` and `INFO("recalc")` are host-dependent by definition, and `CELL("filename",...)` depends on a path that may not exist for a workbook built in memory or loaded from bytes. SheetPort's determinism premise means a formula whose value depends on the host is a hazard, not merely an oddity.

A reasonable shape:

- implement the info-types that are properties of the workbook or the cell rather than the host — `CELL` with `row`, `col`, `width`, `type`, `contents`, `format`, `address`, `prefix`, `protect`; `INFO` with `numfile` and `recalc`
- return `#N/A` or `#VALUE!` for host-dependent selectors rather than inventing a plausible answer, so a workbook cannot silently produce different values on different machines
- if host-dependent selectors are wanted later, gate them behind explicit configuration so the determinism boundary stays visible

`CELL("format",...)` also depends on how much number-format fidelity the workbook layer retains, which is worth checking before committing to it.

drafted with love by (agent) Manfred, with approval from Frankie

PR #329, opened 2026-08-13
feat(workbook): implement CELL/HYPERLINK and fix phantom default shee…

Devider:
## Summary

Implements the `CELL` and `HYPERLINK` builtins in `formualizer-eval`.

### CELL (`crates/formualizer-eval/src/builtins/info.rs`)

Supports the info types `contents`, `address`, `col`, `row` and `type`.

- `address` returns the absolute address and sheet-qualifies cross-sheet
  references (`Sheet2!$A$1`).
- Unsupported Excel info types (`format`, `filename`, `parentheses`,
  `prefix`, `protect`, `width`) and calls without a cell reference return
  `#VALUE!`.
- Only `contents`/`type` read the referenced value; `address`/`col`/`row`
  derive purely from reference metadata, so they never force evaluation of
  the referenced cell or range.
- The function is classified as context-dependent (`WorkbookMetadata`) so the
  formula-plane ingest rejects CELL families like the other metadata builtins.

### HYPERLINK (`crates/formualizer-eval/src/builtins/reference_fns.rs`)

Returns the `friendly_name` when provided, otherwise the `link_location`,
always as text (matching LibreOffice). Errors propagate; multi-cell references
and array constants return `#VALUE!` instead of leaking a debug-formatted
array literal. `_xlfn.`-prefixed spellings resolve through the registry.

## Tests

- `engine/tests/cell_hyperlink_sheet.rs` — 15 cases covering friendly name,
  link-location fallback, `_xlfn.` spelling, error propagation, `#VALUE!` for
  unsupported/missing info types and multi-cell HYPERLINK arguments, plus
  cross-sheet address qualification and whole-column/row addresses.

## Verification

- `cargo test -p formualizer-eval`: 2792 passed, 14 ignored
- `cargo test -p formualizer-workbook --features calamine,umya`: pass
- `cargo fmt --all -- --check` and `cargo clippy -p formualizer-eval --all-targets -- -D warnings`: clean
- `cargo run -p xtask -- schema`: 0 stale files


Devider:
Tests are red. Closing this, will reopen with a fix.

PSU3D0:
Update from our side: your rebase is green now — thanks for staying on it. This is queued for a full review behind the reference-semantics stack currently in flight (#369–#372); it's a substantial PR (CELL/HYPERLINK plus the phantom default-sheet fix) and it overlaps an internal program on default-sheet elimination, so we want to review it against that design rather than piecemeal. Realistic timeline: after the current stack merges. No action needed from you.

PSU3D0 (changes requested):
Requesting changes, with one correction to my own earlier note and a ratification that should make the next round the last one.

Correction first: I previously treated this PR as still carrying the phantom-default-sheet fix. That was wrong on my part — that fix landed via #337 on 2026-08-14, and the current diff contains none of it. Please retitle and rescope the body to CELL/HYPERLINK only so the history doesn't double-count the sheet fix.

Ratification, so you are not left waiting on #239: the direction you chose is accepted as the maintainer position. Deterministic subset only; host-dependent selectors (`filename`, `format`, `color`, etc.) return `#VALUE!`; the missing-reference form returns `#VALUE!` rather than emulating Excel's last-changed-cell behavior, which is host-state-dependent and antithetical to a deterministic engine; omitting `width` for now is fine. Please add a short "Decisions taken" section to the body recording these four choices with a pointer to #239, and we will close #239 against this PR when it merges.

One blocking bug (from #333, measured against this branch): `CELL("address", Sheet2!B7)` returns `$A$1` — it both drops the sheet qualifier and returns the wrong coordinates, meaning the off-sheet reference is not resolved at all. That is independent of any policy question and must be fixed, with tests covering same-sheet, off-sheet, and off-sheet-with-quoted-name references. Expected: `Sheet2!$B$7` with the sheet qualifier present for off-sheet references, per Excel.

Explicitly not blocking: `CELL("contents", blank)` returning `Empty` where Excel returns `0`. That sits inside the open blank-coercion question (#319/#285) and will be adjudicated by the variant study, not patched ad hoc here — keep your current behavior, but add a comment on that test marking the pinned semantics as provisional pending #333, so the study can change it without archaeology.

Two non-blocking notes for a comment or follow-up: all info types call `args[1].value()?` first, so `=CELL("address",A1)` recalcs whenever A1's value changes — over-invalidation only, correctness unaffected, but `address`/`col`/`row` don't need a value dependency; and the `type` doc-comment enumerates b/l/v without saying errors classify as `"v"` (the behavior itself matches Excel). HYPERLINK is clean — `_xlfn.` resolution, error propagation, and the friendly-name number coercion all check out, and the 12-test suite is solid for what it covers.

Fix the address bug, rescope the body, add the decisions section, and this merges.

Devider:
All review feedback addressed:

Scope corrected: PR title and description now focus solely on CELL/HYPERLINK (the phantom default-sheet fix landed separately via #337).

Critical bug fixed: Off-sheet address resolution for CELL("address", Sheet2!B7) now returns Sheet2!$B$7 (was incorrectly returning $A$1). Tests added for same-sheet, off-sheet, and quoted-name references.

Documentation added: A "Decisions taken" section has been added to the PR body, recording the maintainer-approved choices from #239 (deterministic subset only, host-dependent selectors return #VALUE!, etc.).

Provisional test annotated: The test for CELL("contents", blank) is now marked with a comment that its behavior (Empty) is provisional pending resolution of #333.

The branch is rebased onto latest main (conflict resolved), all tests pass (formualizer-eval 2792 passed, workbook tests pass), and fmt/clippy are clean. Ready for review.

PSU3D0:
One addition to my review, from the FormulaPlane span-coverage audit that just completed — it concerns a registration decision in this PR that is invisible in normal testing but load-bearing for the plane cutover.

Today CELL and HYPERLINK reject from formula spans as `unknown_or_custom_function` — the fail-closed path for unregistered names. The moment this PR registers them, their span eligibility is determined entirely by the capabilities/contract you register them with. For CELL the stakes are high: a span broadcasts one template evaluation across many placements, and CELL reads position-dependent properties. If CELL is registered with no context dependence, it becomes span-eligible and `=CELL("row",A2)` dragged down a column would return the anchor cell's answer for every member — a silent wrong-value divergence across the whole family, and nothing in the ordinary test suite would catch it because span promotion needs 100+ member families. Please register CELL with the context-dependence contract (the `ContextDependentFunction` rejection path, same class as ROW/COLUMN) so the plane hard-rejects it. HYPERLINK's scalar text result is span-safe as-is, with one condition: if the link/display side-channel is ever delivered outside the derived-format funnels, it needs the same treatment — for this PR, plain scalar registration is fine.

A regression test pinning "CELL is not plane-eligible" (assert its canonicalization reject) would lock the decision in against future capability refactors.

Devider:
Rebased the branch onto the latest main (a238d067).

The only meaningful conflict was in CHANGELOG.md: both branches had added entries at the top of ### Changed, so I kept both. Everything else merged cleanly.

Re-ran the full verification after the rebase:

formualizer-eval: 2840 passed / 0 failed / 14 ignored
formualizer-workbook with calamine,umya: pass
formualizer-workbook with default features: pass
formualizer-sheetport: pass
cargo fmt, cargo clippy -D warnings: clean
cargo xtask schema: 0 stale files

No behavioral changes were introduced by the rebase. The CELL/HYPERLINK implementation, the WorkbookMetadata context contract, and the regression test ensuring that CELL is not plane-eligible are unchanged.

PSU3D0:
Landed via #397, which carries these commits with original authorship plus six review fixups (reference-error propagation in CELL, 3-D references returning #VALUE!, the generated reference pages, ledger rows, and the contents-on-blank divergence pinned to #333). I could not push to the fork branch from the session I was working in, hence the separate PR. Thank you for the implementation and the quick turnaround on the review asks; CELL and HYPERLINK ship in 0.9.0.

PR #335, opened 2026-08-14
perf(workbook): pre-allocate Arrow capacity in set_values batch path

zhantss:
## Summary

`Workbook::set_values` — the batch path behind `Sheet.setValues` in the JS/WASM
binding — scales super-linearly (≈ O(N²)) on bulk numeric writes. This PR adds a
one-time Arrow pre-allocation, mirroring `set_formulas_inner`, which restores
linear scaling.

## Problem

Each cell written by `set_values_inner` (changelog-off branch) flows through:

```
set_values_inner                              workbook.rs:2574
  → Engine::set_cell_value                     eval.rs:16515
    → mirror_value_to_overlay                  eval.rs:15195
      → ArrowSheet::ensure_row_capacity        arrow_store/mod.rs:3584  (nrows +1 per cell)
        → ColumnChunk::grow_len_to             arrow_store/mod.rs:178   (⚠️ O(old_len))
```

`grow_len_to` does `type_tag.values().to_vec()` — a **full copy of the column's
type-tag array** — plus a full rebuild of every present lane
(numbers/booleans/errors/text) on *every* call. Because `ensure_row_capacity`
grows `nrows` by one per cell, a batch of N cells triggers rebuilds of
O(1), O(2), …, O(N) → **O(N²)** total. The same quadratic path also affects the
changelog-enabled branch (its per-cell `mirror_value_to_overlay`).

## Fix

Pre-allocate the Arrow sheet to the full batch extent once at the top of
`set_values_inner`, exactly as `set_formulas_inner` already does:

```rust
let height = rows.len();
let width = rows.iter().map(|r| r.len()).max().unwrap_or(0);
if height == 0 || width == 0 { return Ok(()); }
let end_row = start_row.saturating_add((height - 1) as u32);
let end_col = start_col.saturating_add((width - 1) as u32);
self.ensure_arrow_sheet_capacity(sheet, end_row as usize, end_col as usize);
```

The per-cell `ensure_row_capacity` then early-returns.

**No behavior change.** `ensure_arrow_sheet_capacity` only grows (never
overwrites) existing Arrow data; `set_value` (single-cell) and
`set_formulas_inner` already call the same helper at the same point. Empty and
ragged batches fall through to the existing early-return / per-cell handling.

## Benchmarks

JS/WASM binding, `setValues`, single numeric column, changelog off:

| rows | before | after | speedup |
|------|--------|-------|---------|
| 1,000 | 16.6 ms | 12.7 ms | 1.3× |
| 5,000 | 121.2 ms | 32.3 ms | 3.8× |
| 10,000 | 415.1 ms | 71.9 ms | 5.8× |
| 20,000 | 1502.5 ms | 164.9 ms | **9.1×** |

Per-cell cost returns to ~constant (~7 µs) instead of growing with N.

## Testing

- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy -p formualizer-workbook --all-targets --all-features -- -D warnings`
- [x] `cargo test -p formualizer-workbook --all-features`

## Checklist

- [x] PR focused and reviewable (13 lines, one private function)
- [x] No public API change (no docs / `public-api` snapshot update needed)
- [x] Conventional commit style (`perf(workbook):`)
- [x] Benchmarks included

### DIFF
--- a/crates/formualizer-workbook/src/workbook.rs
+++ b/crates/formualizer-workbook/src/workbook.rs
@@ -2578,6 +2578,19 @@ impl Workbook {
         start_col: u32,
         rows: &[Vec<LiteralValue>],
     ) -> Result<(), IoError> {
+        // Pre-allocate the Arrow sheet to the full batch extent ONCE, so the
+        // per-cell `mirror_value_to_overlay` → `ensure_row_capacity` → `grow_len_to`
+        // (which rebuilds the whole column's type-tag/lanes on every call) is
+        // amortized to O(N) instead of O(N²). Mirrors set_formulas_inner.
+        let height = rows.len();
+        let width = rows.iter().map(|r| r.len()).max().unwrap_or(0);
+        if height == 0 || width == 0 {
+            return Ok(());
+        }
+        let end_row = start_row.saturating_add((height - 1) as u32);
+        let end_col = start_col.saturating_add((width - 1) as u32);
+        self.ensure_arrow_sheet_capacity(sheet, end_row as usize, end_col as usize);
+
         if self.enable_changelog {
             let sheet_id = self
                 .engine


PSU3D0:
Landed in #336, which carries your commit `e917abad` with your authorship intact. Closing here because the commit was cherry-picked rather than merged, so GitHub cannot detect it automatically — this is landed, not rejected.

Why it went that route rather than a plain merge: the change as written turned `main` red, so merging this branch first would have broken the default branch between the two merges.

## Your optimisation, verified independently

I reproduced it natively rather than trusting the WASM numbers, with `set_formulas` as a null control since it already pre-allocated and therefore must not move:

| rows | set_values before | after | speedup | set_formulas before | after |
|---|---|---|---|---|---|
| 1,000 | 6.8 ms | 4.3 ms | 1.6x | 0.9 ms | 0.8 ms |
| 10,000 | 236.2 ms | 57.3 ms | 4.1x | 9.0 ms | 10.8 ms |
| 20,000 | 834.9 ms | 127.2 ms | **6.6x** | 17.9 ms | 18.4 ms |

Per-cell cost goes from 6.8 us at N=1,000 to 41.7 us at N=20,000 before, and stays flat at 4.3 to 6.4 us after. The null control being unmoved is what makes the speedup attributable to the path you touched. Native gains are smaller than the 9.1x you measured through WASM, which is expected across runtimes; the shape is identical. Your diagnosis of `grow_len_to` rebuilding the type-tag array and every lane per call was exactly right.

I also checked work-equality before trusting any of it, since a faster path that does less work is not an optimisation. Square, ragged, empty, all-rows-empty and offset-anchored batches produce byte-identical observable state: dimensions, `COUNTA`, `COUNTBLANK`, `COUNT`, `SUMPRODUCT`, and the presence or absence of every individual cell.

## The one problem, and it was not really yours

The extent was computed as `rows.len()`. Trailing empty rows write nothing, so a batch ending in an empty row reserved a row past the real extent — and a batch anchored at the last grid row reserved row 1,048,577, which cannot exist.

The existing guardrail `ragged_block_extents_match_mutated_cells` catches this: it asserts `{"rows": 1048576}` and got `1048577`. It lives in `formualizer-cffi`, and your checklist ran `cargo test -p formualizer-workbook`, so it never executed. Worth knowing for next time — `cargo test --workspace` is the one that covers the crate split. That is a sharp edge in our repo rather than an oversight on your part.

**The defect predates your PR.** `set_formulas_inner` has always used that computation and you faithfully mirrored it. Confirmed on plain `main` with nothing applied: `set_formulas` with a trailing empty row reports 1,048,577 rows, and a one-row batch followed by two empty rows reports 3 instead of 1. So reviewing a 13-line performance patch surfaced a latent correctness bug that has shipped in every release. Both call sites are fixed in #336, with 7 tests pinning the extent rules.

Thanks — this was a well-diagnosed, well-scoped, well-benchmarked contribution, and the profile trace in the description made it fast to verify.

zhantss:
@PSU3D0 thank you for the detailed review and for carrying the commit forward in #336. I completely understand the decision to close this PR given the cherry-pick — keeping main green is absolutely the right call.

I really appreciate you taking the time to reproduce the numbers natively and validate the diagnosis. The fact that the native speedup holds up (and that the set_formulas null control stays flat) is reassuring — and your point about the trailing-empty-row bug is a great catch. I'll admit I didn't think to test that edge case, and I'm glad the PR surfaced a latent issue rather than introducing one.

The note about cargo test --workspace vs cargo test -p formualizer-workbook is very helpful — I'll make sure to run the full workspace suite in future contributions. That's definitely a sharp edge worth remembering.

Thanks again for the thorough review, the benchmarks, and the fix in #336 . Looking forward to contributing more in the future!

PR #340, opened 2026-08-16
fix(eval): XLOOKUP Excel-parity no-match and exact-match class rules

tommy230:
## Problem

XLOOKUP had two independent paths that could return a plausible but wrong result.

An omitted `if_not_found` slot, as in `=XLOOKUP(5,B1:B3,C1:C3,,-1)`, evaluated to implicit zero. Excel returns `#N/A`. Exact text needles also crossed value classes, so `"20"` could match numeric `20` through the linear or lowered-text path.

## Solution

Omitted `if_not_found` nodes no longer count as supplied fallbacks. Exact text needles now match only true text cells on both lookup paths and in both search directions.

Per review, the blank-versus-empty-string claim was removed from this PR. Its isolated patch is retained on `pr/xlookup-empty-string-blank-followup` and stays gated on the host-compatibility decision in #319.

## Verification

Rebased onto current upstream `main` at `06dad200fe87cb5bff9c17104c1487cdbf4b4eca`.

```text
cargo fmt --all -- --check                         passed
cargo test -p formualizer-eval xlookup_            16 passed, 0 failed
```

The focused run includes both new regression tests and the existing XLOOKUP tests selected by the filter.

### DIFF
--- a/crates/formualizer-eval/src/builtins/lookup/dynamic.rs
+++ b/crates/formualizer-eval/src/builtins/lookup/dynamic.rs
@@ -458,7 +458,9 @@ impl Function for XLookupFn {
             ));
         }
 
-        if args.len() >= 4 {
+        // An omitted-in-place slot (`XLOOKUP(v,l,r,,mode)`) is not a supplied
+        // if_not_found; Excel returns #N/A, not the omitted slot's 0.
+        if args.len() >= 4 && !args[3].is_omitted() {
             return args[3].value();
         }
         Ok(crate::traits::CalcValue::Scalar(LiteralValue::Error(
--- a/crates/formualizer-eval/src/builtins/lookup/lookup_utils.rs
+++ b/crates/formualizer-eval/src/builtins/lookup/lookup_utils.rs
@@ -112,6 +112,9 @@ impl<'a> PreparedLookupMatcher<'a> {
                 let folded_candidate = candidate_text.to_lowercase();
                 compiled.matches_folded(&folded_candidate)
             }
+            // Excel exact lookups never match a text needle against a
+            // non-text candidate: "20" does not find the number 20.
+            (Some(_), _) => false,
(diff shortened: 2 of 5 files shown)


PSU3D0:
Coming back to this after the release push. Two of the claims here are provable today and we'd like them: the no-match class rules and the exact-match behavior that doesn't involve blank coercion. The third (blank/exact-match coercion) sits in the territory gated by the #319 variant study — same reasoning as on your #319 report: Excel and LibreOffice genuinely disagree in spots here and we don't want to codify one host's quirk before the differential study says which behavior to commit to.

Would you split the PR? Land the two provable claims now (they'll make 0.9.0), and keep the gated claim in a follow-up that we merge as soon as the study gives us ground truth. If splitting is a pain, I can cherry-pick with attribution — your call.

tommy230:
Split complete in `cabd8965`. This PR now contains only the omitted-`if_not_found` fix and the exact text value-class fix, rebased onto current `main` at `06dad200`.

I kept the blank-versus-empty-string change isolated on fork branch `pr/xlookup-empty-string-blank-followup` at `4b80bd88`. I have not opened it upstream because it remains gated on #319.

Focused verification: `cargo fmt --all -- --check` passed; `cargo test -p formualizer-eval xlookup_` passed 16 tests.

PSU3D0:
Re-verified on a merge with current main (post #369/#370/#371): eval lib 2767/0, full workbook suite green, fmt clean. The split is exactly what we asked for — thanks for keeping the blank-vs-empty-string patch staged behind #319 rather than folding it in. Blast radius personally audited: `PreparedLookupMatcher` is consumed only by the X-family in `dynamic.rs`, so the text-never-matches-non-text rule lands precisely where Excel applies it and cannot leak into COUNTIF-style criteria coercion. The lowered-text-lane guard (`get_cell` class check on lane hits) is the right fix for materialized text renderings of numeric cells. Merging.

Issue #330, opened 2026-08-13
bindings: volatile determinism controls (clock, seed, volatile level) are unreachable from the Workbook surface


PSU3D0:
## Summary

`Engine` in Rust exposes a complete runtime control surface for volatile determinism — the clock, the deterministic mode, the workbook seed, and the volatile level can all be changed on a live workbook and take effect on the next recalculation, with no reload. None of it is reachable from the mainstream `Workbook` API in either binding. Only the SheetPort surfaces expose determinism, and only as evaluation options.

## Verified behaviour on main (f85b592d)

Measured through `Engine`, `=RAND()` and `=YEAR(TODAY())`, one engine per volatile level, with a no-change-recalc control to separate a real effect from ambient drift.

| volatile level | control: recalc twice, no change | `set_workbook_seed` then recalc | `set_deterministic_mode` then recalc |
|---|---|---|---|
| `Always` | stable | value changed | 2020 to 2030 |
| `OnRecalc` | drifts, by design | value changed | 2020 to 2030 |
| `OnOpen` | stable | value changed | 2020 to 2030 |

The control column matters: `Always` and `OnOpen` are stable across a no-op recalc, so the observed changes are attributable to the setters rather than to volatiles regenerating anyway.

Mechanically the setters do not need to invalidate anything. Volatile cells are already in the dirty set each evaluation, and the seed and clock are consumed at evaluation time. `Engine::set_workbook_seed` at `engine/eval.rs:3850`, `set_volatile_level` at `:3854`, `set_deterministic_mode` at `:3859`, `set_clock` at `:3877`.

## The gap

`bindings/python/src/workbook.rs` and `bindings/wasm/src/workbook.rs` expose none of these. The `deterministic` parameter present on `register_function` in both is unrelated — it declares whether a user-defined function is cacheable.

Determinism is reachable today only through SheetPort: `evaluate_once(freeze_volatile=, rng_seed=, deterministic_timestamp_utc=, deterministic_timezone=)` in Python at `bindings/python/src/sheetport.rs:252`, and the `freezeVolatile` / `deterministicTimestampUtc` / `deterministicTimezone` keys in wasm at `bindings/wasm/src/sheetport.rs:484`. That is correct for SheetPort, where the call is the session, but it means an ordinary embedder holding a long-lived `Workbook` has no way to pin or advance the clock at all.

## Why this is more than a convenience

[...]

PR #341, opened 2026-08-16
feat(python): deterministic clock setter on Workbook (#330)

tommy230:
## Problem

Python callers holding a live `Workbook` cannot pin its evaluation clock. Deterministic time is available through `SheetPortSession.evaluate_once`, but not through the workbook object used by long-running embedders and Pyodide.

## Solution

`Workbook.set_deterministic_clock` wraps the engine's existing deterministic mode and reuses SheetPort's timezone parser. The new clock applies on the next recalculation without reloading the workbook.

After #357 changed the default temporal egress contract, this branch was rebased onto current `main` and the clock tests now assert native `datetime.date` and `datetime.datetime` values directly.

## Verification

Built the Python extension from the rebased branch with CPython 3.14.6 and `maturin develop --release`.

```text
python -m pytest tests/test_workbook_deterministic_clock.py -q   6 passed
ruff check tests/test_workbook_deterministic_clock.py            passed
ruff format --check tests/test_workbook_deterministic_clock.py   passed
cargo fmt --all -- --check                                      passed
```

### DIFF
--- a/bindings/python/formualizer/formualizer_py.pyi
+++ b/bindings/python/formualizer/formualizer_py.pyi
@@ -1780,6 +1780,18 @@ class Workbook:
             print(wb.evaluate_cell("Data", 3, 1))
         ```
         """
+    def set_deterministic_clock(self, deterministic_timestamp_utc: datetime.datetime, deterministic_timezone: typing.Optional[typing.Any] = None) -> None:
+        r"""
+        Pin the evaluation clock to a caller-supplied instant, so the
+        volatile date/time builtins (TODAY, NOW) evaluate deterministically
+        on the next recalculation. Takes effect on a live workbook; no
+        reload is required.
+        
+        `deterministic_timezone` accepts `"utc"`, `"local"`, or a fixed
+        offset in seconds — the same spelling as
+        `SheetPortSession.evaluate_once(deterministic_timezone=...)`.
+        Omitted means UTC.
+        """
     def evaluate_all(self) -> None: ...
     def last_cycle_telemetry(self) -> CycleTelemetry:
         r"""
--- a/bindings/python/src/sheetport.rs
+++ b/bindings/python/src/sheetport.rs
@@ -288,7 +288,7 @@ impl PySheetPortSession {
     }
 }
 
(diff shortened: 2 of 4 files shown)


PSU3D0:
Reviewed and the implementation is exactly right — reusing `set_deterministic_mode` and the sheetport timezone parser is the shape we wanted for #330, and the clock pins correctly on a live workbook. One thing changed under you though: #357 merged since you opened this, and the engine now egresses native temporal values by default — `=TODAY()` returns a `datetime.date`, not a serial float. Your feature works, but 5 assertions in `test_workbook_deterministic_clock.py` pin the old serial contract and fail against current main (e.g. `assert datetime.date(2026, 3, 14) == 46095.0`).

Two ways to update, either is fine: assert the native types directly (`workbook.evaluate_cell(...) == datetime.date(2026, 3, 14)` — arguably the stronger test now), or call `workbook.set_temporal_egress("serial")` in the fixtures to keep the serial-based assertions as an explicit opt-out. If you rebase onto main and update the pins, this merges immediately — everything else passes on my local run against merged main (133/138, the 5 being these pins). If you'd rather I apply the test modernization on merge, say the word and I'll do it with attribution.

tommy230:
Updated in `4c6aeb0c`: rebased onto current `main` and changed the five stale serial assertions to the native temporal types introduced by #357. The NOW test now pins the full native `datetime.datetime` value.

Verification on the rebased head: `maturin develop --release` passed; the focused test file passed 6 tests; Ruff check and format check passed; Cargo fmt passed.

PSU3D0:
Re-verified on a merge with current main: full Python suite 138/138 on a fresh maturin build, ruff format/check clean, CI green. The modernized pins (native `datetime.date`/`datetime.datetime` per the post-#357 egress contract) are exactly the update we asked for, and reusing `parse_timezone_spec` keeps one timezone spelling across Workbook and SheetPortSession. The live-transition test asserting `live != pinned` before the switch is a nice touch — it proves the setter acts on a live workbook rather than passing by coincidence. Merging; this closes #330.

PR #343, opened 2026-08-18
fix(wasm): harden workbook bindings with coordinate validation

Ocean82:
- Add input validation for cell coordinates in WASM workbook methods
- Throw JS exceptions for invalid coordinates instead of panicking
- Add lock poisoning error handling for sheetNames/getFormula
- Include Python coordinate validation tests for parity

### DIFF
--- a/bindings/python/tests/test_coordinate_validation.py
+++ b/bindings/python/tests/test_coordinate_validation.py
@@ -30,3 +30,34 @@ def test_workbook_cell_apis_reject_zero_based_coords():
         wb.set_values_batch("Sheet1", 0, 1, [[1]])
     with pytest.raises(ValueError, match="1-based"):
         wb.set_formulas_batch("Sheet1", 1, 0, [["=1"]])
+
+
+def test_workbook_cell_apis_reject_out_of_grid_coords():
+    """Coordinates beyond Excel's grid must raise, not abort the interpreter.
+
+    The engine packs coordinates into 20-bit row / 14-bit column fields and
+    asserts on overflow, so these previously aborted the process.
+    """
+    wb = make_workbook()
+
+    too_big_row = 1_048_577
+    too_big_col = 16_385
+
+    with pytest.raises(ValueError, match="exceeds the maximum"):
+        wb.set_value("Sheet1", too_big_row, 1, 99)
+    with pytest.raises(ValueError, match="exceeds the maximum"):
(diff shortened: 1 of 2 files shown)


PSU3D0 (changes requested):
Requesting changes — the crate does not compile. `cargo check --lib -p formualizer-wasm` on your branch fails with E0432 (unresolved imports `crate::workbook::{BindingValue, binding_value}`), E0603 (`reject_unknown_keys` demoted to private but still imported by `inspect.rs:4`), and a knock-on E0282 at `inspect.rs:445`. The PR deletes `BindingValue` and changes visibility in `workbook.rs` but never updates `inspect.rs`; the full crate was evidently never built before pushing. Please build the whole crate, not just the touched module, before requesting review.

On the substance once it compiles: the single-cell and block-write validation is good (including moving validation after marshalling in `setValues`/`setFormulas` so the extent is known), but the coverage is cherry-picked against the PR's own motivation. `Sheet::readRange` (workbook.rs:1522) only goes through `RangeAddress::new`, which checks 1-based/ordering but not the 1,048,576 x 16,384 grid limit — out-of-grid ranges still reach the packed-coord asserts this PR claims to eliminate. `addTable`'s new `validate_table_range` has the same gap: a table spanning rows to `u32::MAX` passes. Either extend the grid bound to ranges and tables or explain why they're exempt.

Two scope items to split out: the ~50-line `BindingValue` removal is a pure refactor unrelated to coordinate hardening (and is exactly what broke the build), and the new tests under `bindings/python/tests/` belong to the Python binding, not a WASM-titled PR. The genuinely good parts — poisoned-lock handling in `sheetNames`/`getFormula` converting silent empties into thrown JS errors, and the explicit grid-limit constant — are worth landing once the above is fixed. Note main has moved substantially since your base (#369–#371, #340–#341); please rebase when you update.

PSU3D0:
Closing with cause. The 2026-08-24 review found the crate does not compile on this branch (`cargo check --lib -p formualizer-wasm`: E0432 unresolved `BindingValue`/`binding_value`, E0603 on `reject_unknown_keys`, and a knock-on E0282 in inspect.rs). There has been no update in the nine days since, while every other PR in the series was resubmitted, so I am reading this one as abandoned rather than in progress.

Coordinate validation and lock-poisoning handling in the WASM workbook bindings are still wanted. If you pick it back up, please open a fresh PR that builds the whole crate and reference this PR for history.

PR #345, opened 2026-08-18
Fix/docs site vuln

Ocean82:


PSU3D0:
Thanks for the bump — the `next` 16.2.6 -> 16.2.11 part is the one we want. But as submitted the docs site fails at install: `fumadocs-mdx@15.0.13` (a major over the previous 14.x) requires a newer `fumadocs-core` than the 16.7.13 pinned here, and its postinstall dies with `ERR_PACKAGE_PATH_NOT_EXPORTED: './content/md/frontmatter' is not defined by "exports" in fumadocs-core/package.json`. Reproduced locally with `pnpm install --frozen-lockfile` on your branch.

Two ways to make this mergeable, either is fine: keep `fumadocs-mdx` on the latest 14.x patch and take only the `next` + `fumadocs-core` patch bumps, or move both fumadocs packages together to a version pair that the fumadocs 15 migration notes say are compatible, and confirm `pnpm install && pnpm build` passes locally. Happy to merge as soon as install+build are clean.

PSU3D0:
One addition from a second audit pass on our side: `docs-site/bun.lock` still pins `next@16.2.6` and the old fumadocs versions, and this PR doesn't touch it — so anyone installing with bun keeps the vulnerable Next.js even after the pnpm lockfile is fixed. When you rework the fumadocs pairing per my previous comment, please either update `bun.lock` to match or delete it so the repo has a single lockfile of record. For context, the `next` bump matters: 16.2.6 is inside the vulnerable range of nine advisories fixed by 16.2.11 (three high, including the Server Actions SSRF), so we do want this PR to land once the install is fixed.

PSU3D0:
Thanks for flagging this. The security bump landed via #391 as a minimal change: next 16.2.6 to 16.2.11 only, which is the first fixed version for all nine advisories affecting 16.2.6 (four high: GHSA-89xv-2m56-2m9x, GHSA-p9j2-gv94-2wf4, GHSA-6gpp-xcg3-4w24, GHSA-m99w-x7hq-7vfj). Two reasons I took it over rather than merging this PR as submitted: it paired the bump with a fumadocs-mdx 14 to 15 major that the peer ranges did not require, and it left docs-site/bun.lock still pinning next 16.2.6, so the vulnerable version stayed resolvable. bun.lock is now deleted since nothing consumes it. Credited you in #391.

PR #346, opened 2026-08-18
Fix/cffi guard hardening

Ocean82:


PSU3D0 (changes requested):
Requesting changes — the advertised fix is not wired in, and the PR's own tests prove it. `guard.rs` (`catch_ffi`, `validate_cell`, `validate_block`, `FZ_MAX_ROW/COL`) has zero call sites anywhere in the crate — no `extern "C"` entry point is wrapped in `catch_unwind`, so despite the module docs, a panic in engine internals still aborts the host process exactly as before. And 3 of your 9 new `ffi_hardening` tests fail on your own branch (`set_cell_value_rejects_zero_based_coordinates`, `set_cell_value_rejects_out_of_grid_coordinates`, `set_values_rejects_block_running_off_the_grid`): they assert `guard.rs`'s message strings ("1-based", "exceeds maximum") while the code path actually taken is main's existing `checked_excel_coordinates`/`validate_cffi_range` ("outside Excel's valid range"). Please run the suite to completion before pushing — `cargo test -p formualizer-cffi --test ffi_hardening` reproduces this immediately.

Required: either wire `catch_ffi` into the public entry points (the actual panic-safety fix the PR advertises, and the part we'd most like to have) or delete `guard.rs`; make the test suite pass against whichever validator is canonical rather than duplicating main's (`validate_block(.., 0, 0)` accepting empty blocks already diverges from main's trailing-trim semantics — two validators will drift). Keep the poisoned-lock hunks exactly as they are: converting ~15 `.unwrap()` lock acquisitions into poisoned-lock error statuses is correct, consistent, and genuinely valuable — that part alone would merge on its own.

Going forward, for hardening PRs we need a call-site demonstration in the body: grep output or an integration test that fails when the guard is removed. Three of your four open hardening PRs add machinery nothing calls, which costs a full review cycle each to discover.

PSU3D0:
Closing with cause. The 2026-08-24 review found that the advertised guard module has no call sites (no `extern "C"` entry point is wrapped, so a panic still aborts the host) and that three of the PR's own nine hardening tests fail on the branch. There has been no update in the nine days since, while every other PR in the series was resubmitted, so I am reading this one as abandoned rather than in progress.

The underlying goal is still wanted. If you pick it back up, please open a fresh PR that wires `catch_ffi` into every `extern "C"` entry point and whose test suite passes on its own head, and reference this PR for history.

PR #347, opened 2026-08-18
fix(eval): Bessel overflow protection and RangeAddress validation

Ocean82:
Bessel functions:
- Add overflow guards for large order values in J/Y Bessel computations
- Prevent infinite loops and NaN propagation for extreme inputs
- Add comprehensive test cases for edge cases

RangeAddress:
- Add validate() method enforcing 1-based and ordered invariants
- Reject 0-based row/col values and inverted ranges
- Add unit tests for validation logic

### DIFF
--- a/crates/formualizer-common/src/range.rs
+++ b/crates/formualizer-common/src/range.rs
@@ -36,21 +36,44 @@ impl RangeAddress {
         })
     }
 
-    pub fn width(&self) -> u32 {
-        self.end_col - self.start_col + 1
+    /// Validate the invariants enforced by [`RangeAddress::new`].
+    ///
+    /// [`RangeAddress`] has public fields and derives `Deserialize`, so values
+    /// reaching the engine from FFI/JSON/CBOR payloads have *not* necessarily
+    /// gone through the checked constructor. Untrusted input should be passed
+    /// through this method before use.
+    pub fn validate(&self) -> Result<(), &'static str> {
+        if self.start_row == 0 || self.start_col == 0 || self.end_row == 0 || self.end_col == 0 {
+            return Err("Row and column indices must be 1-based");
+        }
+        if self.start_row > self.end_row || self.start_col > self.end_col {
+            return Err("Range must be ordered: start <= end");
+        }
+        Ok(())
     }
 
+    /// Number of columns spanned by this range.
+    ///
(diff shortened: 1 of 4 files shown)


PSU3D0 (changes requested):
Requesting changes — the numerics work is real and good, but one behavior regression and one unwired feature block it. The good first: deriving `high_word`/`split_words` from `f64::to_bits()` fixes silently wrong Bessel results on big-endian targets; the `-hx` sign-bit fix, `checked_neg` for `n == i32::MIN`, the Taylor accumulator moving to `f64` (13! already overflows `i32`), and the underflow-safe loop are all legitimate, well-tested fixes we want. The `n > 1000` short-circuit also addresses a genuine DoS — pre-PR, `BESSELJ(2e9, 1)` runs a ~2-billion-iteration recurrence from one formula.

F1 (blocking): that cutoff silently returns wrong values for previously correct inputs. Pre-PR `jn(2000, 100)`/`yn(2000, 100)` computed correctly in a few thousand iterations; post-PR they return `0.0`/`-inf` (surfacing as `#NUM!`). A hardening PR must not regress correct answers. Bound the work instead — cap total recurrence iterations, or use an asymptotic approximation above the threshold — or at minimum document the accuracy contract and justify the constant. Also F4, a tell: `if x.is_infinite() { 0.0 } else { 0.0 }` — both arms identical.

F2 (blocking): `RangeAddress::validate()` has zero non-test callers. The deserialization paths (`#[derive(Deserialize)]` with public fields) are exactly where untrusted ranges bypass `new()`, and none of them call it — the advertised validation is an unused public API. Worse, combined with F3, the saturating `to_sheet_range()` now silently clamps malformed input to row 1 instead of rejecting it, so a bad JSON/CBOR range produces a quietly wrong read rather than an error. Wire `validate()` into the paths that need it (with a test that fails when it's removed), and prefer reject-with-error over silent clamp on untrusted input.

Split suggestion, per the failing-test-first policy for hardening sweeps: the Bessel integer/endianness fixes minus the cutoff would merge almost immediately as their own PR.

PSU3D0:
Closing as superseded by #403, which is the rebased successor of this PR. Review continues there; nothing further is needed here. Thanks for the work.

Issue #303, opened 2026-08-08
[bug] undo of a logged row insert leaves data unrestored and lands the dependency edge one row above the AST reference (wrong values)


PSU3D0:
Found by the AST↔edge invariant harness (branch `test/ast-edge-invariant`, pinned as ignored test `undoing_row_insertion_restores_cross_sheet_ast_and_edge_together`, finding id `AST_EDGE_UNDO_STRUCTURAL_ADJUSTMENT`).

Minimized repro: set `Sheet1!D6 = 'Sheet2'!B6*2`; perform a logged row insertion before row 1 on `Sheet2` — both the retained AST and the graph edge correctly adjust to `Sheet2!B7`; call `undo_logged`. After the undo, the edge points back at `Sheet2!B6` but the retained AST still says `Sheet2!B7`.

This is the most severe of the three findings from this harness because it is not merely a missed dirty: scheduling and evaluation now disagree about which cell the formula reads. The sheet is back in its pre-insertion state, so `B6` is the correct referent — meaning dirty propagation tracks the right cell while evaluation of the formula reads the wrong one (`B7`). The formula can produce actively wrong values while appearing correctly scheduled, and the formula text shown to a user (from the AST) names a different cell than the one whose changes trigger recalculation.

The suspect area is the undo path for structural adjustments restoring graph edge state and AST state through different mechanisms: the edge is restored from a snapshot or inverse shift while the AST's reference adjustment is not inverted alongside it.

The minimized pin will land with the invariant harness; the fix should un-ignore it and make it green. Note the fix likely interacts with the undo-placeholder finding (same undo-engine region); fixing them together with a shared review is probably the right shape.

drafted with love by (agent) Manfred, with approval from Frankie

PSU3D0:
Correction after independent adversarial verification: the mechanism originally described here does not reproduce. In the no-data repro, undo restores AST and edge together (both back to `Sheet2!B6`) — the pinned test for the original claim passes, and this issue as first written was an artifact of an over-minimized repro.

[...]

Issue #301, opened 2026-08-08
[bug] undo of a value write to a referenced empty cell drops the dependent formula edge (silent stale values)


PSU3D0:
Found by the AST↔edge invariant harness (branch `test/ast-edge-invariant`, pinned as ignored test `undoing_value_write_to_referenced_empty_placeholder_preserves_formula_edge`, finding id `AST_EDGE_UNDO_EMPTY_PLACEHOLDER`).

Minimized repro: set `Sheet1!D4 = C6+1` while `C6` is empty, so the graph creates an empty placeholder vertex and an edge from D4 to it. Write a value to `C6` through `Engine::action_with_logger`, then call `undo_logged`. After the undo, D4's retained AST still references C6 but D4 has no outgoing cell edge at all.

Consequence: any subsequent write to C6 will not dirty D4, so D4 serves a silently stale value through every later recalculation. The trigger is an ordinary undo of a value write into a cell that a formula references while empty — a common interactive editing sequence.

The invariant violated is that a formula vertex's outgoing dependency representation must cover every semantic reference in its retained AST. The suspect area is the undo engine's restoration of a placeholder-backed cell: the redo/undo pair appears to remove or fail to restore the placeholder vertex edge rather than returning the graph to its pre-action shape.

The harness that found this asserts AST-derived references (via the T1 `engine::refs` collector) exactly equal graph out-edges plus compressed range dependencies after every operation of randomized edit campaigns, with behavioral dirty-propagation probes as ground truth. The minimized pin will land with the harness; the fix should un-ignore it and make it green.

drafted with love by (agent) Manfred, with approval from Frankie

PSU3D0:
Extension from independent adversarial verification: the redo path has the same defect. After the undo drops the edge, `redo_logged` does not rebuild it either — dependencies remain empty, a subsequent write to C6 does not dirty D4, and D4 evaluates to None. The fix should cover both legs of the undo/redo pair, and the behavioral consequence (stale served value after the undo leg) was independently reproduced against plain main.

drafted with love by (agent) Manfred, with approval from Frankie

PR #348, opened 2026-08-18
test(python): add comprehensive undo/redo/cancel test suite

Ocean82:
Add 23 tests covering:
- value set/overwrite/formula propagation with undo/redo
- multi-undo and redo chains
- begin/end action grouping
- evaluation during undo
- nesting and cancel during eval
- per-workbook isolation
- noop undo/redo
- changelog metadata and set_changelog_enabled
- sheet add/undo and evaluate_cells interaction

### DIFF
--- a/bindings/python/tests/test_undo_redo_cancel.py
+++ b/bindings/python/tests/test_undo_redo_cancel.py
@@ -1,0 +1,349 @@
+"""Tests for undo/redo, begin/end action, and cancel APIs."""
+import formualizer as fz
+
+
+class TestUndoRedo:
+    """Basic undo/redo on single operations."""
+
+    def test_undo_set_value(self):
+        wb = fz.Workbook()
+        s = wb.sheet("S1")
+        s.set_value(1, 1, 42)
+        wb.evaluate_all()
+        assert wb.get_value("S1", 1, 1) == 42.0
+
+        wb.undo()
+        wb.evaluate_all()
+        assert wb.get_value("S1", 1, 1) is None
+
+    def test_redo_set_value(self):
+        wb = fz.Workbook()
+        s = wb.sheet("S1")
+        s.set_value(1, 1, 42)
+        wb.evaluate_all()
+
+        wb.undo()
+        wb.evaluate_all()
+        assert wb.get_value("S1", 1, 1) is None
+
+        wb.redo()
+        wb.evaluate_all()
+        assert wb.get_value("S1", 1, 1) == 42.0
+
+    def test_undo_set_formula_on_existing_cell(self):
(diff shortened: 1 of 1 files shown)


PSU3D0:
Holding this one deliberately: undo/redo currently has two confirmed engine-level defects (#301, #303 — silent stale values and misplaced dependency edges after undo of logged operations), and the next engine tranche rewrites that machinery. Merging a comprehensive test suite against today's behavior would codify the bugs we're about to fix, and then the fix tranche would have to rewrite your tests along with the engine.

Once the undo tranche lands (it's next in the queue after the current review stack), this suite becomes valuable — I'd rather rebase it onto correct semantics than merge it onto broken ones. I'll ping this PR when that happens.

PSU3D0:
Correcting my earlier comment after a deeper audit on our side: I said merging now would codify the undo bugs we're about to fix — that turns out to be inaccurate. We ran your suite against both its base and current main (23/23 green both ways) and checked every assertion against the open undo defects (#301, #303, #367): none of your tests encode buggy behavior as expected. I withdraw that rationale.

The accurate picture: the suite is a solid happy-path regression net, but roughly a third of the tests are assertion-free (e.g. `test_undo_set_formula_on_existing_cell` never asserts C1 after undo despite the comment saying it should be cleared; `test_evaluate_cells_after_undo` assigns a result and asserts nothing; several bare try/except-pass noop tests), and none of the three known undo bugs would be caught by it. So the revised path to merge: tighten or drop the vacuous tests so every test asserts something, rebase off v0.8.4 onto main, and we'll take it ahead of the undo tranche rather than after — the happy-path net is useful while that work happens. The #301/#303/#367 trigger scenarios can come as a follow-up once those are fixed.

PSU3D0:
Thanks, a regression net for undo/redo/cancel is welcome. Before this can merge it needs tightening, because seven of the 23 tests do not assert the behaviour they are named for. Three assert nothing at all: test_cancel_during_evaluate_all, test_undo_without_changes_is_noop and test_redo_without_undo_is_noop. Four more never check the cell their docstring says they verify: test_undo_set_formula_on_existing_cell never reads C1 after the undo, test_sheet_set_formula_undo never reads C3, test_undo_after_sheet_add_delete only asserts the pre-undo state, and test_evaluate_cells_after_undo assigns the post-undo result and never asserts on it. Please either add the assertions or drop those tests, then rebase on main. CI has not run on this PR yet; I will approve the workflow run once the update lands.

PSU3D0:
Closing as superseded by #401, which is the rebased successor of this PR. Review continues there; nothing further is needed here. Thanks for the work.

PR #349, opened 2026-08-18
fix(eval): date arithmetic returns Number, not Date (#312)

### LINKED ISSUE #312: eval: binary +/- preserve a Date type that Excel does not have at the formula level
## Summary

Binary `+` and `-` propagate a `Date`/`DateTime` tag when one operand carries one, so `=A1+1` on a date cell returns a `Date` and `=A1+B1` returns a `Date` whose serial is the sum. Excel has no date type at the formula level: a date cell holds a serial carrying a date number format, arithmetic on it yields a plain number, and the result only *displays* as a date because the cell inherits the source's number format.

This came out of #310, where the tag propagation made every unrepresentable serial an error. That error path is fixed in #311, which was kept deliberately narrow. This issue tracks the underlying design question.

## Observable divergence

With `A1` holding a genuine date value of 2024-12-01 (serial 45627) and `B1 = 45658`:

| Formula | Excel value | Formualizer |
|---|---|---|
| `=A1+B1` | `91285` (number, date-formatted) | `2149-12-04` (Date) |
| `=A1-45627` | `0` (number, date-formatted) | `1899-12-31` (Date) |
| `=A1+0` | `45627` (number) | `45627` as a Date |
| `=A1-B1` | `-31` | `-31` (number, after #311) |

The serials agree throughout; the value *type* does not. Note the internal inconsistency the table exposes: `Date - Date` already yields a number, and after #311 an unrepresentable result yields a number, so the tag survives exactly when the arithmetic happens to land in range. That is a representation artifact, not a rule anyone would choose.

## Why it is not simply a bug to fix

Keeping the tag has real value at the boundaries. It is what lets an ingested date-formatted cell round-trip back to a date on save, and what makes `get_value` hand a Python caller a `datetime.date` instead of a bare float. Dropping it from arithmetic does not threaten that for stored values, but it does change what callers receive from computed cells, and code that reads a computed date column would start seeing floats.

The honest framing is that number format is a presentation concern that Formualizer currently models inside the value domain, and arithmetic is where the conflation becomes visible.

## Options

1. Drop the tag from `+`/`-` entirely. Excel-exact, removes the inconsistency, changes egress types for computed cells.
2. Keep the tag but make it a formatting hint carried alongside the numeric value rather than a distinct variant, so arithmetic is always numeric and display stays date-aware. Larger change, closest to how Excel actually works.
3. Status quo plus #311. Serials are always correct and nothing errors, but computed value types remain divergent from Excel and internally inconsistent.

## Suggested handling

Not a patch-release change: it alters currently-successful results and the value types callers receive, so it wants a minor version and a documented migration note. Worth deciding before 1.0, since the value-type contract is exactly the kind of thing a semver-stable embeddable engine should not revisit afterwards.

Credit to @Devider, whose #310 report identified this as the broader issue behind the error it was reporting.

Ocean82:
Excel has no date type at the formula level. A date cell holds a serial number carrying a date number format; arithmetic on it yields a plain number and the result only displays as a date because the cell inherits the source's number format.

Previously, add_sub_date_aware() reconstructed a Date/DateTime variant when the arithmetic result was representable as a calendar date. This caused observable type divergence from Excel: callers received Date objects instead of numeric serials, and downstream formulas could behave differently depending on whether an intermediate result happened to land in the representable date range.

Now all date arithmetic uniformly returns LiteralValue::Number(serial). The Date/DateTime tags are preserved only in the storage layer for round-trip fidelity on ingested cells; they no longer leak into formula evaluation results.

Closes #312

### DIFF
--- a/crates/formualizer-eval/src/engine/tests/date_arithmetic_negative_serial.rs
+++ b/crates/formualizer-eval/src/engine/tests/date_arithmetic_negative_serial.rs
@@ -3,10 +3,8 @@
 //!
 //! Excel has no date type at the formula level. A date cell holds a serial
 //! number carrying a date number format, so `=A1-B1` with `A1 < B1` is just a
-//! negative number. Formualizer keeps a first-class temporal tag so typed date
-//! values survive a round trip, and propagating that tag through `+`/`-` is a
-//! deliberate convenience -- but it must never turn arithmetic Excel performs
-//! happily into a numeric-domain error.
+//! negative number. All date arithmetic now returns plain `Number` values,
+//! matching Excel's value semantics (#312).
 //!
 //! Each test pins exactly the property it names.
 
@@ -53,14 +51,8 @@ fn date_minus_larger_number_returns_negative_number_not_num_error() {
 #[test]
 fn date_minus_number_at_serial_zero_boundary_stays_representable() {
     let mut engine = engine_with_date_anchor();
-    // Serial 0 is representable as a date, so the temporal tag is preserved.
-    // This pins the boundary the fix must not move.
-    match eval(&mut engine, "=A1-45627") {
-        LiteralValue::Date(d) => {
-            assert_eq!(d, NaiveDate::from_ymd_opt(1899, 12, 31).unwrap())
-        }
-        other => panic!("expected Date at serial 0, got {other:?}"),
-    }
+    // Serial 0 is plain arithmetic: 45627 - 45627 = 0. Excel returns a number.
+    assert_eq!(eval(&mut engine, "=A1-45627"), LiteralValue::Number(0.0));
 }
 
 #[test]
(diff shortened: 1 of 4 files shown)


PSU3D0:
Thanks for digging into #312 — the underlying concern was real, but we ended up resolving it with a different architecture in #357 (merged): the engine now keeps typed temporal values internally and converts at the egress boundary per a `TemporalEgress` policy (`Native` default, `Serial` opt-out), rather than stripping the Date type out of arithmetic. That gives Excel-shaped serials to consumers that want them while keeping native dates at the API surface, and #312 is closed by it.

Stripping the type at the arithmetic level would now conflict with that shipped design, so I'm going to close this one. The effort is appreciated, and the test cases you wrote here informed how we checked the egress work.

PR #351, opened 2026-08-18
fix(eval): MATCH/XLOOKUP exact mode no longer coerces blank to 0 (#319)

### LINKED ISSUE #319: [bug] eval: MATCH exact-match coerces a blank cell to 0, so MATCH(0, range, 0) matches an empty row

Ocean82:
Excel's exact-match lookup functions (MATCH with match_type=0, XLOOKUP, VLOOKUP with range_lookup=FALSE) distinguish blank cells from numeric zero. MATCH(0, {blank,1,2}, 0) returns #N/A in Excel, not position 1.

Previously, find_exact_number_in_view() fell through to find_exact_empty_in_view() when searching for 0, causing blank cells to match as if they contained zero. This affected MATCH, VLOOKUP, HLOOKUP, and XLOOKUP in exact mode.

Remove the blank-equals-zero fallthrough. Searching for Empty still works via find_exact_empty_in_view (called when the needle is Empty).

Closes #319

### DIFF
--- a/crates/formualizer-eval/src/builtins/lookup/lookup_utils.rs
+++ b/crates/formualizer-eval/src/builtins/lookup/lookup_utils.rs
@@ -514,12 +514,8 @@ fn find_exact_number_in_view(
         }
     }
 
-    // Excel-like semantics: Empty cells compare equal to numeric zero.
-    if n.abs() < 1e-12
-        && let Some(idx) = find_exact_empty_in_view(view, vertical)?
-    {
-        return Ok(Some(idx));
-    }
+    // Excel exact-match semantics: blank cells are NOT equal to numeric
+    // zero. MATCH(0, {blank,1,2}, 0) returns #N/A, not a position. (#319)
 
     Ok(None)
 }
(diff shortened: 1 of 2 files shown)


PSU3D0:

PSU3D0:
The #319 oracle study is now ratified (see my comment there), so this is no longer study-blocked. Two of the three changes here match the study: MATCH(0, {blank, ...}, 0) returning #N/A and XLOOKUP(0, ...) falling to the not-found value. The third does not. The test find_exact_number_in_view_does_not_match_blank_as_zero asserts that an Empty needle over {Empty, 1, 2} finds position 1, which is probe S6 in the study, where Excel and LibreOffice both return #N/A. The diff only removes the zero-to-blank fallthrough inside find_exact_number_in_view and leaves the needle dispatch routing LiteralValue::Empty to find_exact_empty_in_view, so a blank needle still matches blank entries. Please make blank range entries unmatchable for every needle in exact mode, keep blank needles coercing to 0 on the needle side (E4 and E8), and replace that assertion with the S6 expectation. CI has not run on this PR yet; I will approve the workflow once the push lands.

PSU3D0:
Closing as superseded by #400, which is the rebased successor of this PR. Review continues there; nothing further is needed here. Thanks for the work.

PR #352, opened 2026-08-19
fix(eval): VLOOKUP/HLOOKUP approximate mode returns #N/A on unsorted …

### LINKED ISSUE #283: [bug] eval: VLOOKUP/HLOOKUP approximate mode has no sortedness guard; behavior contradicts module doc and MATCH

Ocean82:
…data (#283)

VLOOKUP and HLOOKUP with range_lookup=TRUE (approximate mode) now validate that the lookup vector is sorted before binary-searching. Unsorted data returns #N/A instead of silently wrong results.

This makes VLOOKUP/HLOOKUP consistent with MATCH, which already had this guard, and matches LibreOffice Calc behavior. Excel documents approximate results on unsorted data as 'may not be correct'; we choose the safer #N/A to prevent corrupt downstream calculations.

The sortedness check uses the same SearchedVector projection that already handles blank/cross-type entry skipping, so columns with text headers above numeric data or trailing blanks continue to work correctly.

Closes #283

### DIFF
--- a/crates/formualizer-eval/src/builtins/lookup/core.rs
+++ b/crates/formualizer-eval/src/builtins/lookup/core.rs
@@ -3,12 +3,11 @@
 //! Implementation notes:
 //! - MATCH supports match_type: 0 exact, 1 approximate (largest <= lookup), -1 approximate (smallest >= lookup)
 //! - Approximate modes assume data sorted ascending (1) or descending (-1).
-//! - Unsorted-data behavior differs by function and is deliberate: MATCH performs a lightweight
-//!   ascending-order check and returns #N/A when the data is not ordered, while VLOOKUP/HLOOKUP
-//!   bisect without a sortedness guard and can therefore return a row Excel would also return
-//!   incorrectly. Excel documents approximate results on unsorted data as "may not be correct"
-//!   rather than an error, so the unguarded path matches Excel; LibreOffice instead returns #N/A.
-//!   See issue #283 before changing either behavior.
+//! - Unsorted-data behavior: both MATCH and VLOOKUP/HLOOKUP approximate modes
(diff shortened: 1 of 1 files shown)


PSU3D0:

PSU3D0:
Closing as superseded by #399, which is the rebased successor of this PR. Review continues there; nothing further is needed here. Thanks for the work.

PR #353, opened 2026-08-19
fix(common): extend date/time text parsing for Excel compatibility (#…

### LINKED ISSUE #290: [bug] eval: additional date/time text forms rejected that Excel accepts (single-digit year, 24:00, fractional seconds, month-year, datetime inputs to DATEVALUE/TIMEVALUE)
The #289 date-text arithmetic work deliberately scoped itself to the shapes with confirmed corpus impact. An adversarial review probed further and found additional temporal text forms that LibreOffice and Excel accept but Formualizer rejects with `#VALUE!`. None of these are regressions — they fail identically before and after #289 — but they are gaps against the goal of Excel-compatible temporal text handling.

Verified against LibreOffice 24.2.7 headless recalculation (left LibreOffice, right Formualizer):

`="1/1/3"+0` gives `37622` versus `#VALUE!` (single-digit year). `="24:00"+0` gives `1` versus `#VALUE!` (the 24:00 boundary is a full day). `="12:00:00.5"+0` gives `0.500005787` versus `#VALUE!` (fractional seconds). `="Jan 2003"+0` gives `37622` versus `#VALUE!` (month and year with implied day 1). `TIMEVALUE("24:00")` gives `0` versus `#VALUE!`. `TIMEVALUE("12:00:00.5")` gives `0.500005787` versus `#VALUE!`. `TIMEVALUE("1/1/03 12:00")` gives `0.5` versus `#VALUE!` (time extracted from a datetime string). `DATEVALUE("2003-01-01 12:00")` gives `37622` versus `#VALUE!` (date extracted from a datetime string).

Deliberately excluded as a non-goal: `="Jan-03"+0` and `="1/03"+0` return `46025` in LibreOffice because it substitutes the current wall-clock year. Matching that would make results depend on the run date, which conflicts with deterministic evaluation, so these should remain `#VALUE!` and the decision should be documented rather than revisited.

Any work here must extend the centralized parsers in `formualizer-common/src/date_serial.rs` (date-serial ownership is centralized by standing decision), keep the single shared two-digit-year window policy introduced by #289, verify each new accepted form against the oracle before pinning, and add negative tests proving that widening acceptance does not start swallowing genuinely invalid text.

Ocean82:
…290)

Add support for date/time text formats that Excel accepts but Formualizer previously rejected:

- Single-digit year: '1/2/5' parses as January 2, 2005 (2000+n window)
- 24:00 and 24:00:00: accepted as midnight (Excel end-of-day convention)
- Fractional seconds: '12:30:45.5' truncates to '12:30:45' (Excel has no sub-second precision in serial representation)
- Month-year only: 'Jan 2024', 'February 2024' parse as the 1st of that month

All existing date parsing tests continue to pass. The new formats flow through the existing parse_excel_date_text/parse_excel_time_text entry points used by arithmetic coercion, DATEVALUE, and TIMEVALUE.

Closes #290

### DIFF
--- a/crates/formualizer-common/src/date_serial.rs
+++ b/crates/formualizer-common/src/date_serial.rs
@@ -99,6 +99,7 @@ fn parse_numeric_slash_date(text: &str) -> Option<NaiveDate> {
 fn parse_excel_year(text: &str) -> Option<i32> {
     let year = text.parse::<i32>().ok()?;
     match text.len() {
+        1 => Some(2000 + year),
         2 if year <= 29 => Some(2000 + year),
         2 => Some(1900 + year),
         4 => Some(year),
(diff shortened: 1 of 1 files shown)


PSU3D0:

PSU3D0:
Closing as superseded by #398, which is the rebased successor of this PR. Review continues there; nothing further is needed here. Thanks for the work.

PR #363, opened 2026-08-20
fix(workbook): evaluate in-book external references from cached values

Devider:
Excel caches cross-workbook reference values in `xl/externalLinks/externalLinkN.xml` parts and never recalculates them. formualizer used to fail on such references with `#NAME?: Undefined name: [1]Sheet1!A1` at `evaluate_all`, because the workbook adapter never surfaced the cached external-link cells to the engine.

This PR surfaces those cached values so runtime evaluation answers in-book external references exactly as Excel does.

## What changed

- **calamine** (`scan_external_link_sources_from_reader`): scans `workbook.xml` `<externalReferences>` (ordered rIds), maps each `rId -> externalLinkN.xml` via `workbook.xml.rels`, and parses each part's `sheetNames`/`sheetData` into structured `ExternalCachedSource { book_index, sheet, row, col, value }` records. The scan is **deferred to load time** (not at `open_path`/`open_bytes`) so it can be bounded by `WorkbookConfig::ingest_limits`.
- **Structural key matching (F1)**: keys are derived from structured fields — book token + case-folded sheet + 1-based coordinates — via a single shared helper `formualizer_common::external_cell_source_name`, used by both the engine's external-reference resolution (`ExternalReference::source_name()`) and the workbook's seeding. `=[1]Sheet1!A1`, `=[1]Sheet1!$A$1`, `=[1]sheet1!a1`, and quoted punctuated sheet names all resolve to the same key. Raw-text matching is gone.
- **Ingest limits (F2)**: new `WorkbookLoadLimits::max_external_cached_cells` (default 1_000_000) caps the number of cached cells materialized; each part is read with a matching byte budget, so a malicious `externalLinkN.xml` cannot exhaust load-time memory before limits apply.
- **Observability (F3)**: scan failures (zip errors, missing parts/rels) no longer collapse to an empty vec silently — they are counted in `AdapterLoadStats::external_link_scan_failures`, and the seeded count is reported via `AdapterLoadStats::external_cached_source_cells`. Duplicate cached cells (same book/sheet/row/col) are last-wins via the seeding map.
- **Config knob (F3)**: `WorkbookConfig::with_external_cached_sources(bool)` (default **enabled**, matching Excel) restores pre-#363 behavior when disabled.
- **workbook**: `WBResolver` carries the resolved source values and returns them from `resolve_source_scalar` (Name error if a source is missing), wired up in `from_reader_with_adapter_stats`; sources are declared on the engine *before* formula ingest.

Adds Rust (calamine formulas) and Python (openpyxl zip injection) tests covering cached-value evaluation, structural matching (`$`/case/punctuated sheets), the ingest-limit cap, scan-failure recording, the disable knob, and the graceful failure when a workbook references an external book without a cached value.

## Scope

Only **single-cell** references are seeded (`ExternalRefKind::Cell`). Range references like `[1]Sheet1!A1:B2` still route to `resolve_source_table_entry` and fail with `Undefined table`; supporting cached ranges is deferred (tracked separately).

## Policy acknowledgment

The formula plane rejects external references outright (`CanonicalRejectKind::ExternalReference`, `template_canonical.rs`), so this PR makes **runtime evaluation** succeed where template validation refuses. That divergence is deliberate for now; the plane-side follow-up is tracked in #378.

## Decisions taken / seeking ratification

- Cached values are pinned at source version `0` forever — no refresh, no provenance. Acceptable v1 semantics; the staleness model is documented and countable via the new `AdapterLoadStats` fields.
- Key canonicalization case-folds the sheet with Unicode `to_lowercase()` (Excel matches sheet names case-insensitively).
- The external-cache scan reads a bounded byte prefix of each part; truncation seeds fewer cells rather than erroring.
- Duplicate cached cells are last-wins (seeding map).
- `WorkbookLoadLimits` gained a dedicated `max_external_cached_cells` rather than reusing an unrelated sheet cap.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,16 @@ All notable changes to Formualizer will be documented in this file.
 
 ## Unreleased
 
+### Changed
+
+- `WorkbookLoadLimits` is now `#[non_exhaustive]`. Downstream struct-literal construction is disallowed; build one from `WorkbookLoadLimits::default()` and mutate the fields, or use a future builder. This is a deliberate breaking change so limits such as `max_external_cached_cells` can be added later without breaking downstream struct-literals again.
+
+### Fixed
+
+- In-workbook external references (`=[1]Sheet1!A1`) now evaluate from the cached values Excel stores in `xl/externalLinks/externalLinkN.xml` parts (spec §10), instead of failing with `#NAME?: Undefined name`. Excel never recalculates external links, so the cached values are the correct semantics. Sources are keyed structurally (book token + case-folded sheet + 1-based coordinates) via the shared `formualizer_common::external_cell_source_name` builder, so `=[1]Sheet1!A1`, `=[1]Sheet1!$A$1`, and `=[1]sheet1!a1` resolve to the same key. The scan is bounded by `WorkbookLoadLimits::max_external_cached_cells` (default 1,000,000) with a per-part byte budget, scan failures and the seeded count are reported in `AdapterLoadStats` (`external_link_scan_failures` / `external_cached_source_cells`), and `WorkbookConfig::with_external_cached_sources(false)` restores the previous `#NAME?` behavior. Only single-cell references are seeded; range references still route to source tables (`Undefined table`), deferred. (#363)
+
+  This adds public API to `formualizer-common` (`external_cell_source_name`) and `formualizer-parse` (`ExternalReference::source_name`). `formualizer-common` and `formualizer-parse` were published as 3.1.x, so the parse-side addition requires a new parser-track release (3.2.0) before the product crates can depend on it; call that release out when bumping.
+
 ## [0.9.0] - 2026-09-06
 
 ### Highlights
--- a/bindings/python/tests/test_external_links_openpyxl.py
+++ b/bindings/python/tests/test_external_links_openpyxl.py
@@ -6,6 +6,7 @@ Excel stores cross-workbook references inside the same file via
 cached values. formualizer must do the same instead of failing with
 `#NAME?: Undefined name: [1]Sheet1!A1`.
 """
+
 import zipfile
 
 import pytest
@@ -42,18 +43,18 @@ def build_inbook_external_link_xlsx(tmp_path, sheet_data="42", with_cache=True):
     sheet1 = sheet1.replace(
         "</row>",
         f'      <c r="B1">\n        <f t="shared" ref="B1" si="0">[1]Sheet1!A1</f>\n'
-        f'        <v>{sheet_data}</v>\n      </c>\n    </row>',
+        f"        <v>{sheet_data}</v>\n      </c>\n    </row>",
     )
     if with_cache:
         content_types = content_types.replace(
             '<Override PartName="/xl/styles.xml"',
-            f'<Override PartName="/xl/externalLinks/externalLink1.xml" '
-            f'ContentType="application/vnd.openxmlformats-officedocument.spreadsheetml.externalLink+xml"/>\n'
-            f'  <Override PartName="/xl/styles.xml"',
+            '<Override PartName="/xl/externalLinks/externalLink1.xml" '
+            'ContentType="application/vnd.openxmlformats-officedocument.spreadsheetml.externalLink+xml"/>\n'
+            '  <Override PartName="/xl/styles.xml"',
         )
         wbxml = wbxml.replace(
             "</sheets>",
-            "</sheets>\n  <externalReferences>\n    <externalReference r:id=\"rId4\"/>\n  </externalReferences>",
+            '</sheets>\n  <externalReferences>\n    <externalReference r:id="rId4"/>\n  </externalReferences>',
         )
         wbrels = wbrels.replace(
             "</Relationships>",
--- a/bindings/python/tests/test_external_links_openpyxl.py
+++ b/bindings/python/tests/test_external_links_openpyxl.py
@@ -1,0 +1,119 @@
+"""In-workbook external references (spec §10): `[1]Sheet1!A1`.
+
+Excel stores cross-workbook references inside the same file via
+`xl/externalLinks/externalLinkN.xml` parts that cache the referenced values
+(`<sheetData>`); Excel never recalculates external links and surfaces those
+cached values. formualizer must do the same instead of failing with
+`#NAME?: Undefined name: [1]Sheet1!A1`.
+"""
+import zipfile
+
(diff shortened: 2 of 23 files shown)


Devider:
Tests are red. Closing this, will reopen with a fix.

PSU3D0 (changes requested):
Requesting changes — the core mechanism is exactly right (Excel never recalculates external links; surfacing the cached values is the correct semantics) and the tests are well constructed, so we want this. But there is one structural correctness defect, one resource gap, and a policy divergence that need addressing first.

F1 (blocking): key matching is raw-text, exact, case-sensitive. The engine resolves `ReferenceType::External` by exact match on `ext.raw.as_str()` (eval.rs:24017), and the parser stores `raw` as the reference exactly as authored (`raw: reference.to_string()`, parser.rs:864/897) — `$` anchors, case, and interior whitespace included. Your adapter synthesizes keys independently via `format!("[{book_index}]{}!{cols}{row}", quote_sheet(sheet))`. So `=[1]Sheet1!$A$1` silently misses the cache and returns `#NAME?`, as does `=[1]sheet1!A1` (Excel is case-insensitive here), as does any authored spacing; and `quote_sheet` treats `(`/`)`/`\` as safe-unquoted, diverging from parser quoting for punctuated sheet names. Everything hinges on two independently generated strings being byte-identical. The good news: the parser already hands you structured identity — `ExternalBookRef::Token`, `sheet`, and the row/col in `ExternalRefKind` — so please match structurally (book index + case-folded sheet + coordinates) instead of on raw text.

F2 (blocking): unbounded attacker-controlled load. `scan_external_link_sources_from_reader` reads whole XML parts into strings and materializes every cached cell with no cap and no interaction with `WorkbookConfig::ingest_limits` — an xlsx whose externalLink part carries millions of cached rows blows up load-time memory before any limit check runs, while every other ingestion path streams through limits. Wire it into ingest limits (entry cap or byte cap).

F3: scan failures (zip errors, missing parts, malformed rels) collapse to `Vec::new()`, so the workbook "loads successfully" and formulas fail later with a `#NAME?` indistinguishable from a genuinely broken reference. Record scan failures and the seeded-source count in `AdapterLoadStats`; that also resolves F4's visibility half — the staleness model (values pinned at version 0 forever, no refresh, no provenance) is acceptable v1 semantics if it is documented and countable, and a config knob to disable external-cache resolution would complete it.

Smaller items: state the scope explicitly in the body — only single-cell refs are seeded; `[1]Sheet1!A1:B2` still routes to `resolve_source_table_entry` and dies as `Undefined table`, which is fine to defer but not what the title implies. Duplicate synthesized keys currently last-wins silently. And one policy acknowledgment we need in the body: the formula plane rejects external references outright (`CanonicalRejectKind::ExternalReference`), so this PR makes runtime evaluation succeed where template validation refuses — that divergence is deliberate for now, but it must be written down; we will file the plane-side follow-up on our end.

General ask for your PRs going forward, since both of your open ones extend behavior around deliberately-open design questions: a short "Decisions taken / seeking ratification" section in the body lets us ratify in one round instead of discovering the decisions in review.

Devider:
Addressed the review. Reworked the external-link cached-source path:

· F1 structural matching: keys are now derived from structured fields (book token + case-folded sheet + 1-based coords) via a shared external_cell_source_name helper, used by both the engine and the adapter. $/case/quoting variants all resolve to the same key; raw-text matching is gone.
· F2 ingest limits: scan is deferred to load time and bounded by a new WorkbookLoadLimits::max_external_cached_cells cap + per-part byte budget.
· F3 observability: scan failures and the seeded count are recorded in AdapterLoadStats; added WorkbookConfig::with_external_cached_sources(bool) (default enabled).
· Adapter now exposes structured ExternalCachedSource records; duplicates are last-wins.

Also documented scope (single-cell only; ranges still Undefined table), the formula-plane divergence (#378), and a "Decisions taken" section in the PR body.

Tests (Rust + Python), clippy, and public-API snapshots are updated. Ready for final review.

PSU3D0:
Re-reviewed the 2026-08-25 rework. All five requested items are present at 412cfac1: structural identity via external_cell_source_name and ExternalReference::source_name, the max_external_cached_cells ingest limit with the per-part byte budget, the two AdapterLoadStats counters, the resolve_external_cached_sources switch, and the FormulaPlane rejection policy acknowledgment. Thank you for the thorough response; the CHANGES_REQUESTED was stale for a week because nobody re-reviewed, which is on me.

Three things remain before this can merge, and one of them is a real rework rather than a rebase.

First, #386 landed on main today and rewrote the same regions of crates/formualizer-workbook/src/backends/calamine.rs (the CalamineAdapter struct, open_path, open_bytes, open_reader, and the scan_*_from_reader signatures). #386 replaces path/bytes retention with a shared XLSX source exposing SharedXlsxReader::reader(), which is exactly the primitive this PR's ExternalLinkSource { Path, Bytes } retention reimplements. Please rebase and drop ExternalLinkSource in favour of the shared reader; the external-link part scan should open its ZIP part through reader() like the lazy metadata scans do.

Second, WorkbookLoadLimits is not #[non_exhaustive], so adding the public max_external_cached_cells field breaks every downstream struct-literal construction. Either add #[non_exhaustive] to the struct in the same PR (with a note in CHANGELOG since 0.9.0 is a breaking release anyway) or route the limit through a builder method.

Third, the PR adds public API to formualizer-common, formualizer-parse, formualizer-eval and formualizer-workbook and updates four public-api snapshots, but has no CHANGELOG entry. Note that formualizer-common and formualizer-parse were published as 3.1.0 today, so the parse-side additions will need a 3.2.0 parser-track release before the product crates can depend on them; that is fine, just call it out in the entry.

Residual, not blocking: canonical.as_deref().unwrap_or(ext.raw.as_str()) still falls back to raw text for ExternalRefKind::Range, which is the declared out-of-scope path. A one-line comment there saying so would help the next reader.

Once rebased I will re-review promptly. If this misses the 0.9.0 freeze it is a first-class 0.9.1 item, not a stale PR.

Devider:
Rebased onto current main (0.9.0, incl. the #386 shared-XLSX-source work) and reworked per your review: dropped ExternalLinkSource in favor of SharedXlsxReader::reader(), made WorkbookLoadLimits #[non_exhaustive], and added the CHANGELOG entry (with the parser-track 3.2.0 release note). The residual one-line comment on the Range raw-text fallback is in. fmt, clippy, workspace tests, xtask schema, and public-api are all green; Python tests pass including the new openpyxl external-link cases.

PR #369, opened 2026-08-23
fix(eval): IF, IFS, and CHOOSE return selected references

tommy230:
## What was wrong

IF, IFS, and CHOOSE always resolved through their value paths. When one selected a cell or range and another function needed that reference, the reference identity had already been discarded. OFFSET then received a value and returned `#REF!` instead of reading the offset cell.

## Excel behaviour

Measured on Microsoft Excel 16.105.3, Microsoft 365 for Mac, in a synthetic workbook with `A1:A5 = 10..50` and `B1:B5 = 1..5`:

| Formula | Excel | main @ 82a0b0d8 |
| --- | --- | --- |
| `=OFFSET(IF(TRUE,A1,B1),2,0)` | 30 | `Error(Ref)` |
| `=OFFSET(CHOOSE(1,A1,B1),1,0)` (file-load) | 20 | `Error(Ref)` |
| `=OFFSET(IFS(TRUE,A1),1,0)` (replace-entry) | 20 | `Error(Ref)` |
| control `=INDEX(IF(TRUE,A1:A5,B1:B5),3,1)` | 30 | 30 |

IFERROR and IFNA were separately refuted as reference-returning functions: Excel rejected both forms at entry, and legacy-intersection rows confirmed both take their arguments as values. IFERROR also failed replace-based entry, was stripped during file repair, and produced `#VALUE!` when composed through INDEX. This PR is limited to IF, IFS, and CHOOSE.

## The fix

The function contract can now resolve either a reference or a value. Argument handles cache scalar resolution. IF, IFS, and CHOOSE advertise `RETURNS_REFERENCE`, implement reference selection, and retain their existing scalar and array value paths. The repository schema task regenerated all three capability blocks.

This adds the public, doc-hidden `FunctionResolution` enum and provided public `Function::resolve_reference_or_value` method.

## Scope

This branch is standalone on `main`. IFERROR and IFNA are excluded by the Excel oracle. Array-valued selectors retain the accepted two-evaluation residual; the companion INDEX PR pins AST and Arena counts at two. Scalar selectors remain one evaluation.

One temporary seam is explicit: this reference path preserves the original condition error, while the ordinary value path keeps main's `#VALUE!` conversion until the companion IF-error propagation PR merges.

## Verification

The regressions use a separate synthetic engine fixture (a 20-row grid of six value columns) rather than the oracle workbook, so their expected reads are `102`/`402`. All four fail on main with `Error(Ref)` in place of the expected number and pass here. The value-path and scalar selector-count controls pass. Reverting an IF, IFS, or CHOOSE hunk fails its matching regression; all four regressions plus selector coverage pin the shared resolution/cache hunks.

The public-API snapshot was regenerated with `scripts/public-api.sh` on x86_64 Linux under the pinned toolchains and is included as a separate `chore(api)` commit; `scripts/public-api.sh check` exits clean at this tip. The addition is a provided trait method plus its return type, an additive change.

Executed with `rustc 1.97.1`:

```text
cargo clippy --workspace --all-targets -- -D warnings exit 101; same 20 fingerprints as unmodified main
cargo test -p formualizer-eval                      2740 passed; the only failures are the IMEXP/IMSIN/IMCOS cases that fail identically on unmodified main
cargo run -p xtask -- schema --functions IF,IFS,CHOOSE exit 0; 3 matched, 0 stale
cargo run -p xtask -- audit --strict                exit 1; byte-identical base set of 366 missing-example findings, zero naming IF/IFS/CHOOSE
```

No IFERROR/IFNA reference behavior, array-selector optimization, unrelated conditional change, or macOS-generated public snapshot is included.

## How this was found

A workbook exposed OFFSET and INDEX compositions that lost a selected reference. The public regressions reduce that behavior to synthetic addresses and literals.

Drafted by Claude (agent), verified by execution against a real Excel oracle, reviewed by a human before submission.

### DIFF
--- a/crates/formualizer-eval/src/builtins/logical.rs
+++ b/crates/formualizer-eval/src/builtins/logical.rs
@@ -2,7 +2,7 @@
 
 use super::utils::ARG_ANY_ONE;
 use crate::args::ArgSchema;
-use crate::function::Function;
+use crate::function::{Function, FunctionResolution, resolution_to_reference};
 use crate::traits::{ArgumentHandle, FunctionContext};
 use formualizer_common::{ExcelError, LiteralValue};
 use formualizer_macros::func_caps;
@@ -432,7 +432,7 @@ pub struct IfFn;
 /// Variadic: true
 /// Signature: IF(arg1...: any@scalar)
 /// Arg schema: arg1{kinds=any,required=true,shape=scalar,by_ref=false,coercion=None,max=None,repeating=None,default=false}
-/// Caps: PURE, SHORT_CIRCUIT
+/// Caps: PURE, RETURNS_REFERENCE, SHORT_CIRCUIT
 /// [formualizer-docgen:schema:end]
 impl Function for IfFn {
     fn propagate_format(
@@ -442,7 +442,7 @@ impl Function for IfFn {
         result.format_id()
     }
 
-    func_caps!(PURE, SHORT_CIRCUIT, MAY_SPILL);
+    func_caps!(PURE, SHORT_CIRCUIT, RETURNS_REFERENCE, MAY_SPILL);
 
     fn name(&self) -> &'static str {
         "IF"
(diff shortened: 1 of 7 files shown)


PSU3D0:
Full adversarial review complete (probes, LibreOffice cross-oracle, A/B perf, 8 mutants). Verdict: **merging**. Every shipped behavior in your oracle table reproduced exactly, the value-path/reference-path coercion tables survived an 18-selector x 4-nesting divergence attack byte-identical, short-circuit and volatility semantics are clean, and the #372 seam you flagged is fully closed by the merge — all three paths now agree with each other and with LO. Perf: no measurable tax on non-IF workloads (worst delta anywhere +1.6%, inside the noise floor); one structural note, `ArgumentHandle` grew 560 -> 680 bytes from the resolution cache — invisible today, worth watching if more cached fields accrue.

Two follow-up asks from the mutation matrix, non-blocking but real: (1) `resolve_choose_reference_or_value`'s out-of-range bound has zero coverage on either path — widening it by one survives all 2758 tests yet makes `=CHOOSE(3,A1,B1)` panic; please add `CHOOSE(0,...)` / `CHOOSE(len,...)` / `OFFSET(CHOOSE(len,...),0,0)` regressions covering both the value and reference paths. (2) The `may_return_reference()` guard (including its LET/LAMBDA local-name exclusion) is untestable-as-written — a path-taken assertion (your `CountSelectorFn` pattern fits) would make it falsifiable. One data point for your records: LO 24.2 returns 30 for `=OFFSET(IFERROR(A1,B1),2,0)` and `=OFFSET(IFNA(A1,B1),2,0)`, contradicting the Excel measurement behind your IFERROR/IFNA exclusion — your conservative scoping stands, but if you ever re-measure, LO is a vote the other way.

Exceptional work — the refuted-alternatives section and the declared two-evaluation residual made this reviewable at a speed a PR this deep has no right to be.

PR #370, opened 2026-08-23
fix(eval): INDEX reads only its selected cell

tommy230:
> Stacked on #369; the first commit is that PR. Review only the final commit here.

## What was wrong

INDEX materialized an entire bounded rectangle before selecting one element. Runtime dependency tracking therefore recorded every member as read. If an unselected member depended on the INDEX result, the eager rectangle edge produced a circular result even though Excel consumed one cell.

## Excel behaviour vs engine, measured

> **Correction (2026-08-24):** rows K and L below were originally labelled "(control)", implying the pre-fix engine already matched Excel there. Measured on this PR's actual merge base (`ce93ff9e`, i.e. main with #369 merged), both return `#CIRC!` under runtime cycle detection with bulk ingest — the ranges `K1:K3` and `L1:L3` include the INDEX formula's own cell, so the eager rectangle read manufactured a self-edge. They are behavior changes fixed by this PR, not controls. Engine values for rows G, K, and L were measured by execution on 2026-08-24 at `ce93ff9e4dac44e74987ceff06ddd08c1a11ef1b` (base) and `06dad200fe87cb5bff9c17104c1487cdbf4b4eca` (current main) via a harness test in `crates/formualizer-eval` (`cargo test -p formualizer-eval --lib god187v -- --nocapture`, rustc 1.93.0, x86_64 Linux), under `CycleDetection::Runtime` + bulk ingest.

Excel column measured on Microsoft Excel 16.105.3, Microsoft 365 for Mac:

| Formula | Excel | base `ce93ff9e` | main `06dad200` |
| --- | --- | --- | --- |
| G1=7, G2=`=INDEX(G1:G3,1)+1`, G3=9 | 8, no circular flag | `#CIRC!` | 8 |
| K1=`=1/0`, K2=`=INDEX(K1:K3,1)`, K3=9 -> K2 | `#DIV/0!` | `#CIRC!` (behavior change, was mislabelled control) | `#DIV/0!` |
| L1=7, L2=`=INDEX(L1:L3,1)+1`, L3=`=NA()` -> L2 | 8 | `#CIRC!` (behavior change, was mislabelled control) | 8 |
| M1=`=NA()`, M2=`=INDEX(M1:M1,1)` -> M2 | `#N/A` | `#N/A` (control; original claim, not re-measured 2026-08-24) | `#N/A` |
| N1=`=1/0`, N2=`=INDEX(N1:N3,1)`, N3=`=N2` -> N2, N3 | `#DIV/0!`, `#DIV/0!` | circular result (original claim, not re-measured 2026-08-24) | `#DIV/0!` |

The selected error propagates, the unselected error is ignored, and the error-bearing phantom-cycle shape completes acyclically.

## Applicability

The behavior change is only reachable under `CycleDetection::Runtime` combined with the bulk-ingest path. Measured on current main (`06dad200`): with the default `CycleDetection::Static`, G2, K2, and L2 all still return `#CIRC!` — Static stamps every static SCC and no live-edge machinery runs, so default-configured engines are unaffected by this fix. On the incremental path, `set_cell_formula` rejects these self-inclusive INDEX formulas at set time with `#CIRC!` ("Self-reference detected") on both base and main, so the fix is not reachable there for these shapes either.

## The fix

For a bounded reference and one positive in-range selection, INDEX resolves the reference first, computes the selected cell, and materializes only that cell. Selected errors use the same path. Other shapes keep existing value dispatch.

The custom dispatch must reproduce the default dispatch wrapper because it bypasses that wrapper on precise-path success. It therefore applies INDEX's format policy after materialization. `index_precise_path_applies_format_policy` pins that behavior by proving a selected date's source annotation is dropped, matching ordinary INDEX evaluation.

## Scope

Builds on `fix/reference-returning-conditionals`; merge that first. Merge-alone is unsafe because this branch uses its reference-or-value contract.

Bounded cells and omitted dimensions on single vectors use the precise path. Arrays, zero indexes, rectangular omissions, and unbounded `INDEX(Q:Q,n)` keep their existing paths. Static dependencies are unchanged. This commit adds no public item, but inherits the first PR's additions.

## Verification

Six exact tests fail on the stacked base and pass here: four bounded-reference defects, the formula-produced phantom-cycle case, and selector-count fallback coverage. Six controls survive production revert: genuine cycle, selected error, unselected error, format policy, SUM rectangle, and unbounded column. Selector counts are one for zero-index and three fallback shapes, and two for the accepted AST/Arena array residual.

Executed with `rustc 1.97.1`:

```text
cargo clippy --workspace --all-targets -- -D warnings exit 101; same 20 fingerprints as unmodified main
cargo test -p formualizer-eval                      2751 passed; the only failures are the IMEXP/IMSIN/IMCOS cases that fail identically on unmodified main
12 fully qualified exact-test commands              exit 0; one test selected each
```

No static-graph rewrite, unbounded-range precision, array-selector optimization, or public snapshot change is included.

## How this was found

A runtime trace showed INDEX manufacturing edges to unselected cells. Synthetic chains and formula-produced error cells isolate that behavior.

Drafted by Claude (agent), reviewed by a human before submission. Rows G, K, L and the Applicability section were re-measured by execution on 2026-08-24 as noted inline.

### DIFF
--- a/crates/formualizer-eval/src/builtins/logical.rs
+++ b/crates/formualizer-eval/src/builtins/logical.rs
@@ -2,7 +2,7 @@
 
 use super::utils::ARG_ANY_ONE;
 use crate::args::ArgSchema;
-use crate::function::Function;
+use crate::function::{Function, FunctionResolution, resolution_to_reference};
 use crate::traits::{ArgumentHandle, FunctionContext};
 use formualizer_common::{ExcelError, LiteralValue};
 use formualizer_macros::func_caps;
@@ -432,7 +432,7 @@ pub struct IfFn;
 /// Variadic: true
 /// Signature: IF(arg1...: any@scalar)
 /// Arg schema: arg1{kinds=any,required=true,shape=scalar,by_ref=false,coercion=None,max=None,repeating=None,default=false}
-/// Caps: PURE, SHORT_CIRCUIT
+/// Caps: PURE, RETURNS_REFERENCE, SHORT_CIRCUIT
 /// [formualizer-docgen:schema:end]
 impl Function for IfFn {
     fn propagate_format(
@@ -442,7 +442,7 @@ impl Function for IfFn {
         result.format_id()
     }
 
-    func_caps!(PURE, SHORT_CIRCUIT, MAY_SPILL);
+    func_caps!(PURE, SHORT_CIRCUIT, RETURNS_REFERENCE, MAY_SPILL);
 
     fn name(&self) -> &'static str {
         "IF"
@@ -459,6 +459,30 @@ impl Function for IfFn {
         // Single variadic any schema so we can enforce precise 2 or 3 arity inside eval()
         static ONE: LazyLock<Vec<ArgSchema>> = LazyLock::new(|| vec![ArgSchema::any()]);
         &ONE[..]
+    }
+
+    fn eval_reference<'a, 'b, 'c>(
+        &self,
+        args: &'c [ArgumentHandle<'a, 'b>],
+        _ctx: &dyn FunctionContext<'b>,
+    ) -> Option<Result<formualizer_parse::parser::ReferenceType, ExcelError>> {
+        match try_resolve_if_reference_or_value(args) {
+            Ok(Some(result)) => resolution_to_reference(Ok(result)),
+            Ok(None) => None,
+            Err(error) => Some(Err(error)),
+        }
+    }
+
+    fn resolve_reference_or_value<'a, 'b, 'c>(
+        &self,
+        args: &'c [ArgumentHandle<'a, 'b>],
+        _ctx: &dyn FunctionContext<'b>,
+        value_fallback: &dyn Fn() -> Result<crate::traits::CalcValue<'b>, ExcelError>,
+    ) -> Result<FunctionResolution<'b>, ExcelError> {
+        match try_resolve_if_reference_or_value(args)? {
+            Some(result) => Ok(result),
+            None => value_fallback().map(FunctionResolution::Value),
+        }
     }
 
     fn eval<'a, 'b, 'c>(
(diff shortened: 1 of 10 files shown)


PSU3D0:
Full adversarial review complete alongside #369. Verdict: **merging**. The dependency narrowing is the sound shape — static edges stay over-approximate, only live-edge recording narrows — and the sharpest attack we had (moving selector re-closing a genuine cycle: `D1=INDEX(C1:C3,A1)`, `C2=D1+1`, selector walked 1->2->3->2) detected the real cycle and recovered from it correctly in both directions, identically under plane-off and plane-authoritative. Structural ops, #360 named-range interaction, and dirty-edge behavior under edits are all clean. FormulaPlane alignment: neutral by construction — the plane rejects INDEX/IF/IFS/CHOOSE via `FunctionResultSemantics`, and your changes leave every eligibility site untouched, so no divergence is possible.

Three accuracy/verification notes to fix in a fast-follow, none affecting the shipped code: (1) Your oracle table labels the K (selected-error) and L (unselected-NA) rows "(control)" — measured on the merge base both are `#CIRC!` on main, so they're behavior changes this PR delivers, not controls; two of your five fixes are mislabeled as non-fixes. (2) The PR body should name the applicability envelope: the fix is live under `CycleDetection::Runtime` + bulk-ingest load; the default `Static` mode still yields `#CIRC!` for every table row, and `set_cell_formula` rejects these shapes at entry in both modes by existing design. (3) `index_precise_path_applies_format_policy` doesn't pin what the body claims — deleting `apply_format_propagation` from the precise path passes all 2758 tests because INDEX's `propagate_format` is `None`, so the assertion holds vacuously; either make the policy observable in the test or soften the claim. Also from the mutation matrix: `precise_single_cell_selection` (45 lines) is behaviorally indistinguishable from `return true` under the whole suite plus our 212-assertion corpus — the downstream let-chain re-rejects everything it filters. Fine as a perf filter, but a path-taken test would keep future off-by-ones catchable.

PR #377, opened 2026-08-24
test(eval): fast-follow for the #369/#370/#371 review asks; close the all-unbounded OpenRect trap (#376)

### LINKED ISSUE #376: engine: all-unbounded OpenRect key degenerates to a single row-0 stripe in add_range_deps_from_keys (currently unreachable, trap remains)
## Summary

Hardening residual from the #371 review. That PR removed the `Option::zip` that *produced* all-`None` `RangeKey::OpenRect` keys, but the downstream consumer that mishandles such a key is still present: `add_range_deps_from_keys` (`range_deps.rs:818-905`) classifies a four-way-`None` OpenRect as both `col_stripes` and `row_stripes`, falls through both stripe branches, collapses the bounds to `start == end == (0,0)` via the `unwrap_or` chain, and registers a single row-0 stripe — i.e. a fully unbounded range would be tracked as depending on row 1 only.

The state is currently unreachable: the #371 review probed `A:XFD` and `1:1048576` and both parse with at least one bounded axis, and the merged producers always populate at least one bound. But the trap is intact for any future producer of the degenerate key, and its failure mode is the same silent-wrong-value class #371 just fixed.

## Suggested fix

Either a `debug_assert!` in `add_range_deps_from_keys` that at least one axis bound is `Some`, or an explicit branch mapping the all-unbounded key to whole-sheet dependency semantics. Both are small; the assert documents intent, the branch removes the trap entirely.

Surfaced by mutation/edge analysis during the #371 adversarial review (Finding 6); the review report has the exact control-flow walk.

tommy230:
# Fast-follow: test and hardening asks from the #369/#370/#371 reviews, plus #376

Closes #376.

One PR bundling every follow-up the reviews asked for. Each item quotes the ask (short), names the commit, and links to a measured row in the evidence table. Every value in that table was observed by executing the named test at the named commit — nothing is reconstructed.

### From the #371 review

**1. Axis-mirror test** — "add the mirror-shape test — `=SUM($3:$13)` with formula precedents"
`open_rect_row_mirror_sum_pinned` (commit `7dc8ccb3`): the row-band mirror of the shipped column shapes, formula-valued precedents through bulk ingest. Measured failing on the #371 merge base with the exact stale value your review reported (76605 vs 96660) and passing on this branch. Rows E1/E2.

**2. Stripe-precision test** — "a test asserting the exact stripe set ... would pin the precision, not just the presence"
`open_rect_stripe_set_preserves_start_col` (commit `7dc8ccb3`): asserts the consumer of `=VLOOKUP(H3,$D:$F,3,FALSE)` registers exactly column stripes {3,4,5} (0-based) and nothing else. Both safe-over-widening mutants were applied and are killed — dropping `start_col` in the producer (`ingest_builder.rs` `range_key_from_shared`) and in the consumer (`add_range_deps_from_keys`) each widen the observed set to [0..5] and fail the assertion. Rows E3/E4/E5.
One scoping note from verifying the kills: `plan.rs` also builds `RangeKey::OpenRect` values, but that producer feeds only the formula-plane dependency-summary comparison oracle (`dependency_summary.rs`), never stripe registration, so a `start_col` drop there is not observable through scheduling and is outside this test's reach — measured as row E6.

**3. Saturation nit** — "`col + 1` is the one non-saturating conversion in a hunk that saturates everywhere else"
Commit `ef6df6e3`: `cell_ref.coord.col() + 1` → `cell_ref.coord.col().saturating_add(1)` in the eval-plan sample-cell labels.

### From the #369 review

**4. CHOOSE bound regressions** — "please add `CHOOSE(0,...)` / `CHOOSE(len,...)` / `OFFSET(CHOOSE(len,...),0,0)` regressions covering both the value and reference paths"
Commit `d03fd3a4`: `choose_len_boundary_value_path` (direct dispatch, exact `len` and `len+1` boundaries), plus engine-level `choose_selector_bounds_value_path` and `choose_selector_bounds_reference_path` (OFFSET forcing `resolve_choose_reference_or_value`). Your measured mutant — the selector bound widened by one — was applied separately to each path's check: the value-path mutant panics `choose_len_boundary_value_path` with an out-of-bounds argument access, and the reference-path mutant panics `choose_selector_bounds_reference_path` the same way. Rows E7/E8.

**5. `may_return_reference` falsifiability** — "a path-taken assertion (your CountSelectorFn pattern fits) would make it falsifiable"
Commit `d03fd3a4`: we ended up pinning the predicate directly rather than counting calls — `ArgumentHandle` caches its reference resolution, so a CountSelectorFn-style count cannot distinguish the guard from the downstream re-rejection (the same masking that made it untestable end to end). `may_return_reference_syntax_arms` asserts every syntax arm (cell reference, range, RETURNS_REFERENCE function, unbound name, value function, arithmetic, literal), and `may_return_reference_excludes_let_lambda_locals` pins the LET/LAMBDA exclusion both ways via `Interpreter::with_local_env`. Removing the exclusion arm is killed (row E9). Residual: the arena-side arm mirrors the AST arm and is not separately constructible from a unit test; its mutants remain uncovered.

### From the #370 review

**6. Vacuous format-policy test** — "either make the policy observable in the test or soften the claim"
Commit `c75c2fa9`: the precise single-cell fast path is factored out of `IndexFn::dispatch` into `IndexFn::precise_dispatch`, parameterized over the dispatching function. `index_precise_path_applies_format_policy` now drives the same dispatch through a test-only `MarkedIndexFn` whose `propagate_format` is `Some(FormatId(30))` and asserts the marker lands on the result — deleting `apply_format_propagation` from the precise path now fails the test (row E10) instead of holding vacuously. The original INDEX-clears-annotations assertion is kept as a control.

**7. Path-taken test for `precise_single_cell_selection`** — "a path-taken test would keep future off-by-ones catchable"
Commits `c75c2fa9` + `d3120c23`, two layers because we measured the first alone to be insufficient: `index_precise_path_taken_matrix` asserts taken/not-taken through `precise_dispatch` across 11 shapes, but the column bound widened by one *survives* it — `reference_from_base` re-rejects the widened selection, exactly the downstream masking your review described (row E11). `index_precise_selection_predicate_bounds` therefore asserts the filter directly (exposed `pub(crate)`) across the same boundary shapes, and that kills the mutant on the exact widened case (row E12).

### From issue #376

**8. All-unbounded OpenRect key** — "an explicit branch mapping the all-unbounded key to whole-sheet dependency semantics"
Commit `5ff7f933`: both stripe-registration sites (`add_range_deps_from_keys`, and `add_range_dependent_edges`, which accepts arbitrary `SharedRangeRef`s through the public `add_range_edges` and carries the same fall-through) now route a no-bounds-on-either-axis range to `register_whole_sheet_stripes`, registering all 16,384 column stripes. Dirty lookups probe the edited cell's column stripe, so full column coverage reaches the existing precision check, which already treats missing bounds as unbounded. We chose the branch over the `debug_assert!` since it removes the trap rather than documenting it. `all_unbounded_open_rect_key_covers_whole_sheet` constructs the degenerate key directly and asserts both stripe coverage and behavior (an edit at R500C8 dirties the dependent); measured failing with the fix reverted (row E13).

## Measured evidence

Provenance: all runs via `cargo test -p formualizer-eval --lib -- <test>` on the same Linux x86_64 host, rustc 1.93.0. Branch = `fastfollow/review-asks`, tip `d3120c23964a9fd34f9af480ad9f61d51b71088f` (from upstream main `06dad200fe87cb5bff9c17104c1487cdbf4b4eca`); base = #371 merge base `b5bae5a491477d75a1c088ddfd51226438785ece` (mirror test ported to a worktree of that commit as `open_rect_mirror_red.rs`, fixture identical). Mutants were applied to the branch working tree one at a time, the named test executed, output recorded, and the tree restored (verified clean after each run).

| Row | What ran | Where | Observed |
|---|---|---|---|
| E1 | `open_rect_row_mirror_sum_pinned` | base `b5bae5a4` | FAILED — `left: 76605.0, right: 96660.0` |
| E2 | `open_rect_row_mirror_sum_pinned` | branch | ok |
| E3 | `open_rect_stripe_set_preserves_start_col` | branch | ok — set = [3, 4, 5] |
| E4 | same, consumer mutant (`add_range_deps_from_keys`: `sc = 0`) | branch + mutant | FAILED — `left: [0, 1, 2, 3, 4, 5], right: [3, 4, 5]` |
| E5 | same, producer mutant (`ingest_builder.rs`: `start_col: None`) | branch + mutant | FAILED — `left: [0, 1, 2, 3, 4, 5], right: [3, 4, 5]` |
| E6 | same, `plan.rs` OpenRect `start_col: None` probe | branch + mutant | ok — survives; that producer feeds the dependency-summary oracle only (item 2 note) |
| E7 | `choose_len_boundary_value_path`, eval bound `> args.len() - 1` → `> args.len()` | branch + mutant | FAILED — panicked at `choose.rs:173` (out-of-bounds arg access) |
| E8 | `choose_selector_bounds_reference_path`, resolver bound widened the same way | branch + mutant | FAILED — panicked at `choose.rs:204` (out-of-bounds arg access) |
| E9 | `may_return_reference_excludes_let_lambda_locals`, exclusion arm removed | branch + mutant | FAILED — panicked at `functions.rs:995` (local not excluded) |
| E10 | `index_precise_path_applies_format_policy`, `apply_format_propagation` dropped from precise path | branch + mutant | FAILED — `left: None, right: Some(FormatId(30))` |
| E11 | `index_precise_path_taken_matrix`, column bound `col <= cols` → `col <= cols + 1` | branch + mutant | ok — survives; `reference_from_base` re-rejects (motivates E12) |
| E12 | `index_precise_selection_predicate_bounds`, same mutant | branch + mutant | FAILED — `=INDEX(A1:B3,2,3) with dims 3x2: left: true, right: false` |
| E13 | `all_unbounded_open_rect_key_covers_whole_sheet` with `range_deps.rs` reverted to `06dad200` | branch, fix reverted | FAILED — panicked at first stripe assertion (`range_dependencies.rs:1207`) |
| E14 | full `cargo test -p formualizer-eval` (includes the 10 new tests and the rewritten format-policy test) | branch | ok — lib 2775 passed; 0 failed; 14 ignored (main baseline: 2765 passed; 0 failed) |

## Gates

- `cargo fmt --all --check` — clean.
- `cargo clippy --workspace --all-targets -- -D warnings` — clean, zero warnings or errors on rustc/clippy 1.93.0; the pre-existing clippy fingerprints known from other environments did not reproduce on this toolchain, so nothing was suppressed.
- `cargo test -p formualizer-eval` — branch: lib 2775 passed / 0 failed / 14 ignored, all other suites 0 failed. Unmodified main `06dad200` on the same machine: lib 2765 passed / 0 failed, all other suites 0 failed. The IMEXP/IMSIN/IMCOS failures we have seen on other hosts did not occur on this machine on either commit, so no known-base failure list applies to these runs.

drafted by Claude (agent), approved by Tommy

### DIFF
--- a/crates/formualizer-eval/src/builtins/lookup/choose.rs
+++ b/crates/formualizer-eval/src/builtins/lookup/choose.rs
@@ -651,6 +651,44 @@ mod tests {
     }
 
     #[test]
+    fn choose_len_boundary_value_path() {
+        let wb = TestWorkbook::new().with_function(Arc::new(ChooseFn));
+        let ctx = wb.interpreter();
+        let f = ctx.context.get_function("", "CHOOSE").unwrap();
+
+        let a = lit(LiteralValue::Text("A".into()));
+        let b = lit(LiteralValue::Text("B".into()));
+
+        // CHOOSE(len, ...): the last choice is the highest valid index.
+        let two = lit(LiteralValue::Int(2));
+        let args = vec![
+            ArgumentHandle::new(&two, &ctx),
+            ArgumentHandle::new(&a, &ctx),
+            ArgumentHandle::new(&b, &ctx),
+        ];
+        let result = f
+            .dispatch(&args, &ctx.function_context(None))
+            .unwrap()
+            .into_literal();
+        assert_eq!(result, LiteralValue::Text("B".into()));
+
+        // CHOOSE(len + 1, ...): exactly one past the last choice must be
+        // #VALUE!, not an out-of-bounds argument access. A bound widened by
+        // one panics here while every wider overflow still errors.
+        let three = lit(LiteralValue::Int(3));
+        let args = vec![
+            ArgumentHandle::new(&three, &ctx),
+            ArgumentHandle::new(&a, &ctx),
+            ArgumentHandle::new(&b, &ctx),
+        ];
+        let result = f
+            .dispatch(&args, &ctx.function_context(None))
+            .unwrap()
+            .into_literal();
+        assert!(matches!(result, LiteralValue::Error(e) if e.kind == ExcelErrorKind::Value));
+    }
+
+    #[test]
     fn choose_decimal_index() {
         let wb = TestWorkbook::new().with_function(Arc::new(ChooseFn));
         let ctx = wb.interpreter();
--- a/crates/formualizer-eval/src/builtins/reference_fns.rs
+++ b/crates/formualizer-eval/src/builtins/reference_fns.rs
@@ -165,7 +165,7 @@ impl IndexFn {
         }
     }
 
-    fn precise_single_cell_selection<'a, 'b>(
+    pub(crate) fn precise_single_cell_selection<'a, 'b>(
         args: &[ArgumentHandle<'a, 'b>],
         rows: u32,
         cols: u32,
--- a/crates/formualizer-eval/src/builtins/reference_fns.rs
+++ b/crates/formualizer-eval/src/builtins/reference_fns.rs
@@ -290,6 +290,38 @@ impl IndexFn {
         }
     }
 
+    /// Attempt the precise single-cell fast path.
+    ///
+    /// Returns `None` when any gate declines, in which case the caller falls
+    /// back to `validated_dispatch`. Factored out of `dispatch` (and
+    /// parameterized over the dispatching function) so tests can assert which
+    /// path a given argument shape takes and observe the format policy being
+    /// applied to the precise result.
+    pub(crate) fn precise_dispatch<'a, 'b, 'c>(
+        function: &dyn Function,
+        args: &'c [ArgumentHandle<'a, 'b>],
+        ctx: &dyn FunctionContext<'b>,
+    ) -> Option<crate::traits::CalcValue<'b>> {
+        let argument = args.first()?;
+        if !argument.may_return_reference() {
+            return None;
+        }
+        let Ok(FunctionResolution::Reference(base)) = argument.resolve_reference_or_value() else {
+            return None;
+        };
+        let (rows, cols) = Self::bounded_dimensions(&base)?;
+        if !Self::precise_single_cell_selection(args, rows, cols) {
+            return None;
+        }
+        let Some(Ok(reference @ ReferenceType::Cell { .. })) =
+            Self::reference_from_base(args, ctx, base)
+        else {
+            return None;
+        };
+        let value = Self::materialize_reference(ctx, &reference).ok()?;
+        Some(function.apply_format_propagation(value))
+    }
+
     fn validated_dispatch<'a, 'b>(
         &self,
         args: &[ArgumentHandle<'a, 'b>],
@@ -393,16 +425,8 @@ impl Function for IndexFn {
         args: &'c [ArgumentHandle<'a, 'b>],
         ctx: &dyn FunctionContext<'b>,
     ) -> Result<crate::traits::CalcValue<'b>, ExcelError> {
-        if let Some(argument) = args.first()
-            && argument.may_return_reference()
-            && let Ok(FunctionResolution::Reference(base)) = argument.resolve_reference_or_value()
-            && let Some((rows, cols)) = Self::bounded_dimensions(&base)
-            && Self::precise_single_cell_selection(args, rows, cols)
-            && let Some(Ok(reference @ ReferenceType::Cell { .. })) =
-                Self::reference_from_base(args, ctx, base)
-            && let Ok(value) = Self::materialize_reference(ctx, &reference)
-        {
-            return Ok(self.apply_format_propagation(value));
+        if let Some(value) = Self::precise_dispatch(self, args, ctx) {
+            return Ok(value);
         }
         self.validated_dispatch(args, ctx)
     }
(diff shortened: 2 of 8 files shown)


PSU3D0:
Reviewed and merging. Verified on a merge with current main (post #340/#341): eval lib 2777/0, fmt clean, and I spot-checked evidence row E4 personally — the consumer mutant fails `open_rect_stripe_set_preserves_start_col` with exactly the `[0..5]` vs `[3,4,5]` output your table records, and the unmutated baseline passes. The provenance discipline in this body — every value tied to a named test at a named commit, residuals declared (E6, E11, the arena-side arm) rather than papered over — is exactly the contract we asked for on #371, delivered in full. Thank you for turning it around this fast.

Two notes for the record, neither blocking: choosing the explicit whole-sheet branch over the `debug_assert!` for #376 was the right call (it removes the trap instead of documenting it), and guarding `add_range_dependent_edges` as well — the second site accepting arbitrary `SharedRangeRef`s through public `add_range_edges`, which our review had not flagged — makes the fix complete rather than symptomatic. The `precise_dispatch` factoring is a faithful extraction (the old let-chain's error fallthrough maps to `.ok()?` with the same `validated_dispatch` fallback), and parameterizing it over the dispatching function is a clean way to make the format policy observable without a test-only hook in production code.

This closes out every follow-up from the #369/#370/#371 reviews. drafted with love by (agent) Manfred, with approval from Frankie

PR #384, opened 2026-08-25
feat(eval): write plane derived formats to the Arrow computed-overlay lane

PSU3D0:
Stacked on #380 (format broadcast parity) — merge that first.

## Problem

Derived number-formats lived in two stores: the engine side-band map and the Arrow computed-overlay format lane. Legacy vertex evaluation dual-writes both, but plane flushes only recorded the side-band, so the plane structurally diverged from legacy and cutover could not rely on lane-first egress for span output.

## Fix

Option<FormatId> now threads through ComputedWriteBuffer / SpanComputedWriteSink entries and survives last-write-wins coalescing as part of the winning entry. Every flush shape (point, sparse, dense, run) applies formats in one chunk-local batched pass after value application; rect writes carry a parallel format matrix. Side-band recording is retained (intentional dual-write during transition). Unformatted results write None, clearing stale lane entries. No public API changes.

## Tests

The three broadcast oracles now also assert overlay-lane FormatIds after flush, plus a proof test that a plane-flushed date cell egresses Date with its side-band entry manually cleared (lane-first resolution).

## Validation

cargo test -p formualizer-eval --lib: 2787 passed. computed_flush 24 passed. clippy -D warnings; fmt --check.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,24 @@ Formualizer combines broad Excel-compatible formula support with Arrow-backed st
 - **SaaS products** embedding spreadsheet logic — pricing calculators, planning tools, configurators — without shipping a full spreadsheet UI.
 - **Data engineers** extracting business logic trapped in spreadsheets into reproducible, testable pipelines.
 
+## Building an AI agent? Use agent-spreadsheet
+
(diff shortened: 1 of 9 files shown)


PSU3D0:
Closing: this branch's head commit is already an ancestor of main (merged via the stacked merge commits e2c2bf14 / e2ee3cf1). Nothing left to land here.

PR #385, opened 2026-08-25
feat(formula-plane): admit cell/scalar-arm IF/IFS/CHOOSE as spans (reject-then-revoke)

PSU3D0:
Stacked on #380 (format broadcast parity) — merge that first.

## What

IF, IFS, and CHOOSE become FormulaPlane span-admissible when every result/reference-role arm is a finite single-cell reference or a statically scalar expression. Relative, absolute, mixed-anchor, and cross-sheet arms; nested admitted calls; scalar aggregate arms such as SUM(A1:A3).

## How

The canonicalizer rejects these functions doubly today (ReferenceReturningFunction and ArrayOrSpillFunction caps). After argument canonicalization a bottom-up shape pass revokes both reasons per function only when every occurrence is proven safe and scalar-result-shaped; reject-then-revoke can never widen admission. The same rule is mirrored into the arena-native canonical-label path and the read-projection builder's duplicate gate, so production ingest matches tree canonicalization. Dependency summaries needed no changes: Value-context recursion already emits cond-union-then-union-else precedents per contract 8.6/9.4.

## Firewall (load-bearing)

Range arms stay legacy in v1 by construction: multi-cell range arms spill on legacy but would truncate to top-left through span publication. Pinned goldens with legacy-vs-authoritative equality: IF(TRUE,A1:A3,0), CHOOSE(2,A1,B1:B3), SQRT(IF(...)), SUM(IF(...)) = 600. Still-rejected sweep asserts unchanged reject strings for range, open-range, named, volatile, OFFSET/INDIRECT/INDEX, and colon-operator shapes.

## Tests

New formula_plane_ref_returning_admission.rs (546 lines): parity fill-downs, short-circuit untaken-error pin, error-code parity (#VALUE! conditions, IFS #N/A, CHOOSE selector errors), memo group-merge incl. format parity on equal-values-different-formats, taken/untaken arm dirty propagation, guarded self-reference InternalDependency fallback, conditional-cycle CycleMember demotion convergence. Coverage pins updated for the newly admitted population.

## Validation

cargo test -p formualizer-eval --lib: 2797 passed at commit, green post style fix. clippy -D warnings; fmt --check. CHANGELOG entry included. Contract section 9.2 amendment text drafted for the internal design bundle (reference_returning_static_unimplemented no longer describes admitted static shapes).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ All notable changes to Formualizer will be documented in this file.
 
 ### Changed
 
+- FormulaPlane now admits `IF`, `IFS`, and `CHOOSE` spans when every result arm is a single cell or a statically scalar expression. Range, open-range, named, volatile, and dynamic-reference arms remain on the legacy path.
 - Named-range definitions now track structural inserts and deletes like all other references; previously they stayed pinned to their original coordinates. Deleting a band containing a name target now invalidates the definition to `#REF!`. (#170)
 - Computed temporals are numeric during evaluation (`ISNUMBER`/`TYPE` now match Excel). Native scalar, range, table, Python, and SheetPort egress materializes date/time values from the cell's effective format; callers can opt into uniform raw serials. A known datetime class preserves midnight datetimes, while calamine's code-lossy date-ish signal can only classify pure fractions as time, day-plus-fraction serials as datetime, and integers as date.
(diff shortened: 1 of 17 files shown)


PSU3D0:
Closing: this branch's head commit is already an ancestor of main (merged via the stacked merge commits e2c2bf14 / e2ee3cf1). Nothing left to land here.

PR #398, opened 2026-09-02
fix(common): extend date/time text parsing for Excel compatibility (partially addresses #290)

### LINKED ISSUE #290: [bug] eval: additional date/time text forms rejected that Excel accepts (single-digit year, 24:00, fractional seconds, month-year, datetime inputs to DATEVALUE/TIMEVALUE)

Ocean82:
## Summary

Extends date/time text parsing in `formualizer-common` for closer Excel compatibility (single-digit years, `24:00`/`24:00:00` as midnight, truncated fractional seconds, and `MMM YYYY` / `Month YYYY` forms resolving to the first of the month).

This partially addresses #290 and supersedes #353, rebased onto current `main`.

## Review follow-ups (this revision)

- `parse_month_year_only` now requires a four-digit year, so `"Jan 3"` is no longer read as `2003-01-01`. This removes the inconsistency with the already-rejected month-plus-day-without-year forms (`"Jan-03"`, `"1/03"`).
- `strip_fractional_seconds` only strips a dot that terminates a full `HH:MM:SS` field (two colons before the dot), so a single-colon `"12:00.5"` is rejected as `#VALUE!` rather than silently accepted as `12:00`. Negative test added.
- Eval-layer regressions in `engine/tests/date_text_arithmetic.rs` pin `="Jan-03"+0`, `="1/03"+0`, `="Jan 3"+0` and `="12:00.5"+0` as `#VALUE!`.

## Scope / non-goals

This does not close #290. Three expectations require a design decision and are split into #416:

- `="24:00"+0` must be `1` (a full day) while `TIMEVALUE("24:00")` stays `0`.
- Sub-second retention (`="12:00:00.5"+0` = 0.500005787).
- `DATEVALUE`/`TIMEVALUE` routing through the datetime parser.

## Changes

- `crates/formualizer-common/src/date_serial.rs`: the two parser fixes plus unit tests.
- `crates/formualizer-eval/src/engine/tests/date_text_arithmetic.rs`: eval-layer #VALUE! regressions.

## Testing

- `cargo test -p formualizer-common date_serial` — 17 passed
- `cargo test -p formualizer-eval --lib date_text_arithmetic` — 7 passed


PSU3D0 (changes requested):
Thanks for rebasing. I ran the full #290 expectation table through the formula path on main and on your head. Single-digit years and the month-year forms are genuine wins (`="1/1/3"+0`, `="1/2/5"+0`, `="Jan 2003"+0`, `="January 2003"+0`, case-insensitive), `TIMEVALUE("24:00")` is 0 as required, common and eval suites are green, clippy and fmt are clean, and nothing in the parity corpus moves.

Requesting changes, because the branch meets three of the eight #290 expectations, breaches the issue's one explicit non-goal, and says "Closes #290".

Blocking, all parser-local and small.

`"Jan 3"` now parses as 2003-01-01. `parse_month_year_only` goes through the shared `parse_excel_year`, which you extended to accept one-digit years. #290 explicitly refuses month-plus-day-without-year forms (`"Jan-03"`, `"1/03"`) because LibreOffice fills in the wall-clock year, and your branch correctly still rejects those two, so this is internally inconsistent as well. Gate the year token in `parse_month_year_only` to four digits, or two and four; I would take four only, since `"Mar 05"` reading as March 2005 rather than March 5 is a coin flip against Excel.

`strip_fractional_seconds` truncates at the first dot anywhere, so `"12:00.5"` is accepted as 12:00 by both coercion and TIMEVALUE. Restrict it to a dot that follows an HH:MM:SS field (two colons before it), and add a negative test.

Add eval-layer regressions for the non-goals in engine/tests/date_text_arithmetic.rs: `="Jan-03"+0`, `="1/03"+0`, `="Jan 3"+0`, `="12:00.5"+0` all `#VALUE!`. Nothing at that layer covers them today, which is why the breach passed CI.

Not met and not parser-local, so please retitle to "partially addresses #290" rather than closing it: `="24:00"+0` must be 1 (a full day) while `TIMEVALUE("24:00")` stays 0, which a `NaiveTime`-returning parser cannot express; fractional seconds must be kept (`="12:00:00.5"+0` is 0.500005787, your branch gives 0.5), which conflicts with the deliberate whole-second contract in `time_to_fraction`; and `DATEVALUE("2003-01-01 12:00")` and `TIMEVALUE("1/1/03 12:00")` are still `#VALUE!` because those functions never route through `parse_excel_datetime_text`. I split those three into #416 so they get a design decision before anyone implements them.

Separate and pre-existing, not yours: `DATEVALUE("13/1/2003")` returns 37634 on main via a silent day/month swap; filed as #415.

PSU3D0:
Rechecked the updated head: the four-digit month-year gate, fractional-minute rejection, and requested formula-layer negative regressions now address our earlier blockers. Focused common/date_serial and eval/date_text_arithmetic tests and formatting pass; hosted CI is green. Merging this deliberately partial improvement; #290/#415/#416 are not being declared complete. Thank you for making the requested corrections.

PR #399, opened 2026-09-02
fix(eval): VLOOKUP/HLOOKUP approximate mode returns #N/A on unsorted data (#283)

### LINKED ISSUE #283: [bug] eval: VLOOKUP/HLOOKUP approximate mode has no sortedness guard; behavior contradicts module doc and MATCH

Ocean82:
## Summary

VLOOKUP/HLOOKUP in approximate mode now return `#N/A` on unsorted data instead of a misleading match, matching Excel behavior. Closes #283.

This supersedes #352, rebased cleanly onto current `main`.

## Changes

- `crates/formualizer-eval/src/builtins/lookup/core.rs`: unsorted-data guard for approximate VLOOKUP/HLOOKUP, with tests.

## Testing

- `cargo test -p formualizer-eval --lib lookup` — 210 passed, 0 failed

## Notes

- Single-file, focused change. Conventional commit style.


PSU3D0 (changes requested):
Thanks for rebasing. Policy first, because #283 was still open: we are taking the guard. Excel documents its approximate result on unsorted data as possibly incorrect, which is a licence to return anything, so bit-exact parity with an undefined result is not a parity property worth defending, and our consumers are pipelines where a silent wrong row is the worst outcome. I have recorded that on #283. The performance concern I had going in does not apply either: the approximate arm already clones the whole column per call, so the guard measured 1.02 to 1.04 times on 200k-row tables with 2k and 20k lookups (the underlying per-call clone is filed as #414).

Verified on your head: `VLOOKUP(2.5, {3,"c";2,"b";1,"a"}, 2, TRUE)` and the HLOOKUP equivalent return `#N/A` on both the reference and array-literal paths; the text-header-plus-trailing-blanks case still resolves; mixed-type columns pass because the projection filters to the needle's type class; `match_type = -1` is untouched. Lookup tests (210), full eval suite (2864), clippy and fmt are green. No bench scenario or fixture uses approximate lookups on unsorted data, so no checksums move.

Requesting changes for three things, all small.

The function-level docstrings contradict the new behaviour and they feed the generated site pages. core.rs:135 (MATCH) still says "see the module notes for how VLOOKUP/HLOOKUP differ", pointing at the module note this commit rewrote; core.rs:460 (VLOOKUP) and core.rs:719 (HLOOKUP) still say unsorted or descending data can return incorrect rows. Please fix all three and regenerate docs-site/content/docs/reference/functions/lookup/vlookup.mdx and hlookup.mdx.

The MATCH array-literal branch (core.rs:392 to 396) calls the unguarded `binary_search_match`, on main and on your head. After this PR, `VLOOKUP(2.5, {3,"c";2,"b";1,"a"}, 2, TRUE)` is `#N/A` while `MATCH(2.5, {3,2,1}, 1)` returns 3, which is a new inconsistency created by a change whose purpose is consistency. Please extend the guard to that branch.

Add the tests Copilot asked for: HLOOKUP approximate on unsorted data, both reference and array-literal paths, and the MATCH array-literal case.

Optional: fold the guard into `binary_search_match` behind a flag rather than keeping a near-duplicate that builds the `SearchedVector` projection a second time. LOOKUP stays unguarded for now and is called out on #283 as the remaining decision.

PSU3D0:
@Ocean82 We are preparing the 0.9.0 release candidate and keeping this ratified behavior change on the release path. I am checking your latest head against the outstanding review, including the MATCH literal branch, generated documentation and horizontal tests. We will prepare any remaining follow-up on a separate integration branch, preserving your original commits and authorship rather than rewriting your branch. If you are actively updating this, please let us know by 2026-09-06 14:00 UTC so we can incorporate your work and avoid duplication. We will keep unrelated lookup performance work separate from this semantic change.

PSU3D0:
The tested completion head is now available in draft #438 (fe01f09f), preserving both of your original commits and authorship. The shared ascending/descending guard, literal/reference paths, original-position mapping and generated docs pass independent review and focused/full Rust validation; fresh combined-main CI is running. The safer unsorted-input policy is documented without claiming Excel always returns an error.

The September 6 14:00 UTC coordination checkpoint is unchanged. Please send any active follow-up or feedback there; your branch has not been overwritten.

PSU3D0:

PSU3D0:
Integrated through #438 with your original commits/authorship preserved and the reviewed follow-up fixes. The final X-function correction also passed the contributor’s corrected captured-Excel rerun. Closing this superseded PR rather than merging the older head; thank you for the contribution.

PR #400, opened 2026-09-02
fix(eval): MATCH/XLOOKUP exact mode no longer coerces blank to 0 (#319)

### LINKED ISSUE #319: [bug] eval: MATCH exact-match coerces a blank cell to 0, so MATCH(0, range, 0) matches an empty row

Ocean82:
## Summary

MATCH/XLOOKUP in exact mode no longer coerce a blank cell to numeric `0`. `MATCH(0, {blank,1,2}, 0)` now returns `#N/A` rather than matching the blank, matching Excel exact-match semantics. Closes #319.

This supersedes #351, rebased cleanly onto current `main` (the lookup_utils.rs changes since the original base merged automatically and were verified by hand).

## Changes

- `crates/formualizer-eval/src/builtins/lookup/lookup_utils.rs`: remove the empty==0 shortcut in `find_exact_number_in_view`; add a regression test asserting blank is not matched as 0 while `Empty` needles still find blanks.
- `crates/formualizer-eval/src/engine/tests/dynamic_lookup_arrow.rs`: test expectation update.

## Testing

- `cargo test -p formualizer-eval --lib lookup` — 209 passed, 0 failed

## Notes

- Focused change. Conventional commit style.

### DIFF
--- a/crates/formualizer-eval/src/builtins/lookup/lookup_utils.rs
+++ b/crates/formualizer-eval/src/builtins/lookup/lookup_utils.rs
@@ -517,12 +517,8 @@ fn find_exact_number_in_view(
         }
     }
 
-    // Excel-like semantics: Empty cells compare equal to numeric zero.
-    if n.abs() < 1e-12
-        && let Some(idx) = find_exact_empty_in_view(view, vertical)?
-    {
-        return Ok(Some(idx));
-    }
+    // Excel exact-match semantics: blank cells are NOT equal to numeric
+    // zero. MATCH(0, {blank,1,2}, 0) returns #N/A, not a position. (#319)
 
     Ok(None)
 }
@@ -806,6 +802,62 @@ mod tests {
         );
         assert_eq!(
             find_exact_index_in_view(&view, &wildcard, true, DateSystem::Excel1900).unwrap(),
+            Some(1)
+        );
+    }
+
+    #[test]
+    fn find_exact_number_in_view_does_not_match_blank_as_zero() {
+        // Regression test for #319: MATCH(0, {blank, 1, 2}, 0) must return
+        // None (which surfaces as #N/A), not match the blank cell.
+        let values = vec![
+            LiteralValue::Empty,
+            LiteralValue::Number(1.0),
+            LiteralValue::Number(2.0),
+        ];
+        let engine = build_vertical_text_engine(&values, 8);
+        let range = ReferenceType::range(
+            Some("Sheet1".to_string()),
+            Some(1),
+            Some(1),
+            Some(3),
+            Some(1),
+        );
+        let view = engine.resolve_range_view(&range, "Sheet1").unwrap();
+
+        // Searching for 0 must NOT match the blank cell at index 0.
+        assert_eq!(
+            find_exact_index_in_view(
+                &view,
+                &LiteralValue::Number(0.0),
+                false,
+                DateSystem::Excel1900
+            )
+            .unwrap(),
+            None
+        );
+        assert_eq!(
+            find_exact_index_in_view(&view, &LiteralValue::Int(0), false, DateSystem::Excel1900)
+                .unwrap(),
+            None
+        );
+
+        // Searching for Empty should still find the blank cell.
+        assert_eq!(
+            find_exact_index_in_view(&view, &LiteralValue::Empty, false, DateSystem::Excel1900)
+                .unwrap(),
+            Some(0)
+        );
+
+        // Searching for 1 still works normally.
+        assert_eq!(
+            find_exact_index_in_view(
+                &view,
+                &LiteralValue::Number(1.0),
+                false,
+                DateSystem::Excel1900
+            )
+            .unwrap(),
             Some(1)
         );
     }
--- a/crates/formualizer-eval/src/engine/tests/dynamic_lookup_arrow.rs
+++ b/crates/formualizer-eval/src/engine/tests/dynamic_lookup_arrow.rs
@@ -13,8 +13,9 @@ fn xlookup_whole_column_empty_lookup_matches_first_cell() {
         .set_cell_value("Sheet1", 1, 2, LiteralValue::Int(42))
         .unwrap();
 
-    // Lookup column A has no used rows; XLOOKUP should still be able to resolve A:A without
-    // materializing the entire million-row range.
+    // Lookup column A has no used rows; XLOOKUP(0,...) should NOT match blank
+    // cells because Excel's exact match distinguishes blank from zero (#319).
+    // The "not found" value "NF" is returned.
     engine
         .set_cell_formula("Sheet1", 1, 3, parse("=XLOOKUP(0,A:A,B:B,\"NF\")").unwrap())
         .unwrap();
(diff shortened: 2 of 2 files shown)


tommy230:
Thanks for rebasing this. I reran the ratified #319 boundary against `7af70402`. The forward numeric-needle case is fixed, but the blank-needle and reverse exact paths still select a blank entry.

Fixture: A1=1, A2 blank and never written, A3 as shown; F1 blank and never written.

| probe | A3 | expected | PR head |
|---|---:|---:|---:|
| `=MATCH(F1,A1:A3,0)` | 2 | `#N/A` | 2 |
| `=MATCH(F1,A1:A3,0)` | 0 | 3 | 2 |
| `=XLOOKUP(0,A1:A3,A1:A3,"NF",0,-1)` | 2 | `"NF"` | 0 |

The first two pin the asymmetry ratified on #319: a blank range entry is never an exact-match candidate, while a blank needle still coerces to numeric zero and therefore continues to a real zero. The current `LiteralValue::Empty => find_exact_empty_in_view(...)` dispatch implements S6 in the opposite direction. The reverse XLOOKUP path compares through `PreparedLookupMatcher`, where `cmp_for_lookup` still makes an Empty candidate equal to zero.

The warmed lookup cache should be covered too. `LookupIndex::find_first_exact` and `find_last_exact` still fall back from a numeric-zero needle to `first_empty`, so the corrected result needs to remain stable after the index builds.

Could this PR make Empty candidates unmatchable across every exact path, preserve blank-needle-to-real-zero coercion, and replace the Empty-needle assertion with the ratified S6 expectation? Criteria matching remains out of scope, as agreed on #319.

authored by Codex (agent), with approval from Tommy

PSU3D0 (changes requested):
Thanks for rebasing. I reran the ratified #319 boundary independently and I can confirm every row in tommy230's table above, plus a few more. Requesting changes, because this is still the numeric-needle half of the fix, and the rebase introduced a new inconsistency that main did not have.

What is fixed on your head: `MATCH(0, …, 0)`, `XMATCH(0, …, 0)`, exact `VLOOKUP`/`HLOOKUP` with a 0 needle, forward `XLOOKUP(0, …)` and its wildcard mode all correctly refuse a blank entry. What is not: a blank needle still selects a blank entry (`MATCH(F1, A1:A3, 0)` returns 2 whether A3 is 2 or 0, where the study says `#N/A` and 3), reverse `XLOOKUP(…, -1)` still returns 0 for both a 0 needle and a blank needle, and the horizontal blank-needle case is the same.

The new problem is the warmed lookup index. `LookupIndex` builds once a range has at least 64 cells and the same key has been looked up more than three times. On your head, `MATCH(0, A1:A3, 0)` over {1, blank, 2} gives `#N/A`, but the identical layout over A1:A100 (with a sentinel in A100) gives 2 after the index builds, and stays 2 on the next recalc. So the same formula now answers differently cold and warm; main was at least uniformly wrong. The same flip applies to the VLOOKUP, HLOOKUP, XMATCH and forward XLOOKUP rows that look fixed on a 3-row fixture.

Paths that still need the change, all in crates/formualizer-eval/src:

builtins/lookup/lookup_utils.rs:476, the `LiteralValue::Empty => find_exact_empty_in_view(...)` dispatch. This implements S6 backwards. Route an Empty needle into `find_exact_number_in_view(view, 0.0, …)`; `find_exact_empty_in_view` becomes dead.

builtins/lookup/lookup_utils.rs:110 to 120, `PreparedLookupMatcher::matches`. The non-text arm falls through to `cmp_for_lookup`, which via `value_to_f64_lenient` treats an Empty candidate as 0. Return false for an Empty candidate before that fallback; the needle side keeps coercing to 0. This single edit fixes both XLOOKUP reverse scans and both fallback scans in dynamic.rs (lines 348, 362, 690, 703).

engine/lookup_index_cache.rs: delete the `first_empty` fallback in `find_first_exact` (189 to 193) and `find_last_exact` (206 to 210), make `LookupHashKey::from_literal` return None for Empty cells so blanks are never indexed as candidates, map an Empty needle to `Number(0.0)` at the query sites, and remove the now-unused `first_empty` field and `numeric_zero_candidate`.

Your test `find_exact_number_in_view_does_not_match_blank_as_zero` (lookup_utils.rs:845 to 850) still asserts that an Empty needle over {Empty, 1, 2} finds position 1, which is probe S6 where Excel and LibreOffice both return `#N/A`. Please flip it to `None` and add a companion asserting an Empty needle finds a real 0 cell. Then add engine-level regressions for the blank-needle, reverse-XLOOKUP and horizontal cases both cold (3 rows) and warm (100 rows, several formulas over the same range), since the warm path is what let this survive two rounds.

It is about 40 lines across three files. This is on the 0.9.0 path; if you would rather I take it from here with your commit as the base, say so and I will keep your authorship on the numeric-needle half.

tommy230:
Two days quiet here. From my side I am happy for you to take it from Ocean82's commit as the base with their authorship preserved on the numeric-needle half, if they are. The blank-needle, reverse-XLOOKUP and warm-index rows are the ones I would want covered cold and warm. Happy to re-run the ratified boundary against whatever head lands.

PSU3D0:
@Ocean82 @tommy230 We are preparing the 0.9.0 release candidate and will complete the ratified blank-candidate/blank-needle asymmetry, reverse scans and warmed-index cases on top of Ocean82's original contribution, preserving its authorship. I will prepare the remaining changes on a separate integration branch and link a tested head for the requested independent boundary rerun. If Ocean82 is actively updating this, please let us know by 2026-09-06 14:00 UTC so we can incorporate that work without competing branch updates. Criteria matching remains out of scope.

PSU3D0:
@Ocean82 @tommy230 The tested completion head requested for the independent boundary rerun is now draft #438 (fe01f09f). Ocean82’s original contribution and authorship are preserved. Blank candidates are excluded; blank needles retain numeric-zero behavior, including signed zero. Adversarial no-zero and false/text-zero-after-zero cases now cover materialized/reverse paths as well as actual warm forward index service; reverse scans are not mislabeled as cache hits. Criteria matching is unchanged.

Independent final review and focused/full Rust validation pass; fresh PR CI is running. The September 6 14:00 UTC checkpoint remains unchanged, and feedback or a fresh oracle rerun is welcome before integration.

PSU3D0:

PSU3D0:
Integrated through #438 with your original commits/authorship preserved and the reviewed follow-up fixes. The final X-function correction also passed the contributor’s corrected captured-Excel rerun. Closing this superseded PR rather than merging the older head; thank you for the contribution.

PR #402, opened 2026-09-02
fix(docs-site): bump fumadocs-mdx 14->15 to fix js-yaml vuln

Ocean82:
## Summary

Bumps `fumadocs-mdx` from `14.2.11` to `15.0.13` in `docs-site` to drop the transitive `js-yaml` dependency flagged by SNYK-JS-JSYAML-18313070 (high severity).

This supersedes #345. The original branch conflicted because the docs dependency tree moved on `main`; this version reapplies the fix cleanly on current `main`.

## Changes

- `docs-site/package.json`: `fumadocs-mdx` 14.2.11 -> 15.0.13. `fumadocs-core`/`fumadocs-ui` left at `16.7.7` to match the current tree.
- `docs-site/pnpm-lock.yaml`: regenerated with `pnpm install --lockfile-only`.

## Testing

- `pnpm install --lockfile-only` — resolves cleanly
- `pnpm peers check` — no peer dependency issues
- `js-yaml` now resolves to `5.4.1` only in the lockfile (the vulnerable version is gone)

## Notes

- Keeping `fumadocs-core`/`ui` at 16.7.7 avoids the peer-dependency mismatch that a core bump would introduce.

### DIFF
--- a/docs-site/package.json
+++ b/docs-site/package.json
@@ -15,7 +15,7 @@
   "dependencies": {
     "formualizer": "^0.6.0",
     "fumadocs-core": "16.7.7",
-    "fumadocs-mdx": "14.2.11",
+    "fumadocs-mdx": "15.0.13",
     "fumadocs-ui": "16.7.7",
     "lucide-react": "^0.570.0",
     "next": "16.2.11",
--- a/docs-site/pnpm-lock.yaml
+++ b/docs-site/pnpm-lock.yaml
@@ -15,8 +15,8 @@ importers:
         specifier: 16.7.7
         version: 16.7.7(@mdx-js/mdx@3.1.1)(@types/estree-jsx@1.0.5)(@types/hast@3.0.4)(@types/mdast@4.0.4)(@types/react@19.2.17)(lucide-react@0.570.0(react@19.2.7))(next@16.2.11(react-dom@19.2.7(react@19.2.7))(react@19.2.7))(react-dom@19.2.7(react@19.2.7))(react@19.2.7)(zod@4.4.3)
       fumadocs-mdx:
-        specifier: 14.2.11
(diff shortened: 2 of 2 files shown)


PSU3D0 (changes requested):
Thanks for reapplying this on current main. The security goal is genuinely met: `pnpm audit` on your head no longer reports the js-yaml advisory (4.3.0 → 5.4.1) and the esbuild low goes away too. One correction for the commit message: fumadocs-mdx 15 does not drop js-yaml, it still depends on it directly, just at a non-vulnerable version.

Requesting changes because the docs site does not build on this head, and main does. Both `pnpm install --frozen-lockfile` (at the postinstall step) and `pnpm build` fail with:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find module '.../node_modules/fumadocs-core/dist/content/md/frontmatter.js'
  imported from '.../node_modules/fumadocs-mdx/dist/index-file-C9Bw7He1.js'
```

fumadocs-mdx 15 deep-imports that path, and fumadocs-core 16.7.7 does not ship it (it has `dist/content/md.js` as a file, not a directory); 16.8.0 and every later 16.x do. The declared peer range on fumadocs-mdx@15.0.13 is `^16.7.0`, which is too loose, so `pnpm peers check` passes while the runtime breaks. Nothing else about the 14 → 15 move affects this repo: source.config.ts, next.config.mjs and the `fumadocs-mdx:collections/server` import are unchanged.

The fix is to bump `fumadocs-core` and `fumadocs-ui` alongside, to any 16.8+. I verified `16.15.4` with `fumadocs-mdx` kept at `15.0.13`: install, full `pnpm build` and the 466-page static export all succeed, and two pre-existing image-size highs disappear as a bonus. Please do not jump to fumadocs-mdx 15.4 in the same PR; it peers on core ^16.15.3 plus new satteri packages, a larger change than this needs. Regenerate the lockfile after the bump.

Nothing in CI builds docs-site, which is why every check is green here; I filed #413 for that.

PSU3D0:
I completed the narrow dependency alignment requested in the earlier review: core/ui are now 16.15.4, retaining mdx 15.0.13, with a regenerated lockfile (933b22e2). Your original contribution is preserved.

Verified the MDX generator, full production build and 466-file Markdown export. The local pnpm 11 default install generated MDX but then rejected ignored third-party build scripts (esbuild/sharp); I did not approve those scripts or change trust settings. Frozen-lockfile installation with scripts disabled, explicit project MDX generation, and the production build succeeded. The audit no longer reports js-yaml, but still reports 2 critical, 6 high and 3 moderate advisories elsewhere; this is not an all-dependencies security cleanup. The change upgrades js-yaml rather than removing it. Docs CI remains separately tracked in #413.

PSU3D0 (approved):
Core/UI alignment resolves the documented MDX import failure. Production docs build/export passed locally, and updated-head CI is green. Remaining unrelated advisories and the pnpm script-policy qualification are documented.

PR #403, opened 2026-09-02
fix(eval): endian-safe Bessel word extraction and integer-overflow guards

Ocean82:
## Summary

Endian-safe Bessel word extraction and integer-overflow guards for `BESSELJ`/`BESSELY`. This is the split, review-addressed successor to #347 — it keeps only the endianness/integer-safety fixes and resolves the blocking feedback.

## What changed vs. #347 (addresses review)

- **F1 (blocking regression) resolved.** The `n > 1000` short-circuit that silently returned `0.0`/`-inf` for still-representable orders (e.g. `jn(2000,100)`, `yn(2000,100)`) is removed. Replaced with a documented `MAX_RECURRENCE_ORDER` (200_000) iteration cap that returns the asymptotic limit **only** for orders that have provably underflowed (J -> 0) or diverged (Y -> +/-inf). Every representable result runs the real recurrence unchanged. Accuracy contract is documented on the constant.
- **F4 (dead branch) resolved.** The `if x.is_infinite() { 0.0 } else { 0.0 }` arm lived inside the removed cutoff and is gone.
- **F2 (RangeAddress::validate) excluded.** Per the split suggestion, this PR contains no `RangeAddress` changes — only the Bessel fixes.

## Retained (the wanted fixes)

- `high_word`/`split_words` derived from `f64::to_bits()` (fixes silently-wrong results on big-endian targets).
- `hx ^= i32::MIN` sign-bit flip instead of `-hx`.
- `checked_neg().unwrap_or(i32::MAX)` for `n == i32::MIN` in `jn`/`yn`.
- `f64` Taylor accumulator (13! overflows i32).
- i64-widened backward-recurrence loop; exhaustive `n & 3` match without unreachable arms.

## Testing

- `cargo test -p formualizer-eval --lib bessel` — 19 passed, 0 failed
- Includes endianness/bit-layout, extreme-order and negative-zero panic-safety, and an accuracy-contract regression (moderate orders run the real recurrence; only pathological orders short-circuit).

## Notes

- Rebased onto current `main`.


PSU3D0 (changes requested):
Thanks for the split; this is the right shape. Every retained fix is present and correct, and I verified the two that matter most empirically: main panics on `jn(i32::MIN, 1.0)` (negate with overflow) and on `jn(2_000_000_000, 1.0)` in debug (add with overflow; 19 s in release), and the branch handles both. The #347 blocker is gone: `jn(2000, 100)` and `yn(2000, 100)` now run the real recurrence and agree with scipy exactly. Full eval suite (2833), clippy and fmt are green.

Requesting changes on the replacement bound, because Copilot's objection holds numerically. `asymptotic_large_order` at bessel_jn_yn.rs:79 decides on `n` alone, but whether `J_n(x)` has underflowed depends on `n` relative to `|x|`. The first order that underflows is 518 at x=100, about 11,770 at x=1e4, about 1.0e6 at x=1e6 and about 1.0e8 at x=1e8. Against scipy: `jn(300000, 1e8)` is 2.6539906e-5 and the branch returns 0.0; `yn(300000, 1e8)` is -7.5245326e-5 and the branch returns -inf, so `=BESSELY(1e8,300000)` becomes `#NUM!`; `jn(250000, 250000)` is 7.1005611e-3 and the branch returns 0.0; and there is a discontinuity of 7.8e-6 between n=200000 and n=200001 at x=1e8. Main computes all of these correctly in about 2 ms. The PR's own `bessel_pathological_order_short_circuits_to_limit` test locks the wrong behaviour in, and the doc comment on `MAX_RECURRENCE_ORDER` (the claim that every representable order is far below the cap, and the `n ln(2n/(e x))` formula) is only valid for n much larger than x.

Concrete fix: make the short-circuit depend on x using the Debye underflow criterion, which is the actual boundary and reuses the 709.78 constant the file already has.

```rust
fn asymptotic_large_order(n: i32, x: f64, kind: BesselKind) -> Option<f64> {
    let n_f = n as f64;
    let ax = x.abs();
    if n_f <= ax { return None; }              // below the turning point: always compute
    let a = ax / n_f;                           // 0 < a < 1
    let s = (1.0 - a * a).sqrt();
    let xi = ((1.0 + s) / a).ln() - s;          // J_n ~ exp(-n*xi) / sqrt(2*pi*n)
    if n_f * xi <= 709.8 { return None; }       // still representable
    Some(match kind { BesselKind::J => 0.0, BesselKind::Y => f64::NEG_INFINITY })
}
```

Checked against the measured thresholds: at x=1e8 this short-circuits from n≈1.00038e8 (scipy's first zero is 100,037,790), lets `jn(300000, 1e8)`, `jn(250000, 250000)` and `jn(1e6, 1e8)` fall through to the recurrence, and still resolves `jn(2e9, 1.0)` in O(1) so the DoS stays fixed. For the residual `n <= |x|` case (for example `jn(2e9, 3e9)`, genuinely finite, about 3 s on main) add a separate iteration cap chosen for latency and return `f64::NAN` rather than `0.0`; engineering.rs:2417 already maps non-finite results to `#NUM!`, which is an honest answer where a confident zero is not.

Please also update the short-circuit test, add positive regressions for the three values above plus a continuity check across n=200000/200001 at x=1e8, fix the doc comment, and add a `### Fixed` CHANGELOG line. Minor and non-blocking: the J short-circuit returns +0.0 before `sign` is computed, so odd n with negative x loses the sign of zero.

PSU3D0:
Parking this implementation rather than merging a new numerical regression. Compiling the exact head shows Y_100(0.06) = -infinity, while an 80-digit reference is -5.76410997427483e307 (finite); J_100(0.06) is also prematurely zeroed. The recurrence cap also intercepts the existing infinity shortcut. The argument-dependent criterion is an improvement over the earlier order-only guard, but its shared threshold still does not prove J underflow/Y overflow. The valuable byte/integer safety work, exact reproductions and replacement acceptance criteria are preserved in #465, including contributor attribution. Thank you for the work; closing this mixed approach does not mean the underlying safety fixes are unwanted.

PR #408, opened 2026-09-02
fix: bound parser recursion depth to prevent stack overflow on deeply nested formulas

chiliec:
## What

The recursive-descent expression parser (`parse_bp`) has no depth limit. A formula string with deep nesting recurses one stack frame per level until the thread stack overflows and the **process aborts** (`fatal runtime error: stack overflow, aborting`).

`parse()` takes an untrusted formula string — an embeddable spreadsheet engine ingests formulas from arbitrary workbooks / user input — so this is a denial of service. A ~1 KB string reliably crashes the host process.

Reproduced (before this change) on an 8 MB main-thread stack:

```rust
let f = format!("={}1{}", "(".repeat(1000), ")".repeat(1000)); // =((((...1...))))
let _ = formualizer_parse::parser::parse(&f); // -> stack overflow, aborts the process
```

Same for a long unary-minus chain (`=-----…1`) and nested calls (`=SUM(SUM(SUM(…)))`).

## Fix

Track recursion depth on the `Parser` and return a `ParserError` once it exceeds `MAX_PARSE_DEPTH` (64, matching Excel's own formula-nesting limit) instead of recursing further. `parse_bp` becomes a thin wrapper that increments/decrements the counter around the existing logic (renamed `parse_bp_inner`), so the guard covers every nesting path (parens, unary, infix, calls).

Normal formulas are unaffected — 40-deep nesting still parses, and all 321 existing parser tests pass.

## Test

Adds a regression test that parses hostile deeply nested input (parens / unary / calls, 5000 deep) on a small **1 MiB** thread stack and asserts an `Err` is returned rather than the process aborting. On `main` this test aborts; with the fix it passes.

`cargo test -p formualizer-parse` (322 pass), `cargo clippy --workspace`, and `cargo fmt --all --check` are all clean.

---
*Disclosure: prepared with AI assistance; the crash was reproduced and the fix compiled/tested locally before submitting.*

### DIFF
--- a/crates/formualizer-parse/src/parser.rs
+++ b/crates/formualizer-parse/src/parser.rs
@@ -2324,7 +2324,17 @@ pub struct Parser {
     /// When > 0, treat a top-level `OpInfix(",")` as a terminator (call-arg
     /// separator) instead of the union/list operator. Used by `parse_call_arguments`.
     in_call_args_depth: usize,
+    /// Current recursion depth of `parse_bp`. Bounded by [`MAX_PARSE_DEPTH`] so a
+    /// deeply nested formula (e.g. `=((((...))))`) returns an error instead of
+    /// overflowing the stack.
+    depth: usize,
 }
+
+/// Maximum expression nesting depth accepted by the parser. Excel itself caps
+/// formula nesting at 64; matching that keeps every real formula well within
+/// bounds while preventing unbounded recursion (and stack overflow) on hostile
+/// input.
+const MAX_PARSE_DEPTH: usize = 64;
 
 impl Parser {
     /// Tokenize a formula using the default Excel dialect and prepare it for parsing.
@@ -2372,6 +2382,7 @@ impl Parser {
             volatility_classifier: None,
             dialect,
             in_call_args_depth: 0,
+            depth: 0,
         }
     }
 
@@ -2508,6 +2519,22 @@ impl Parser {
     }
 
     fn parse_bp(&mut self, min_precedence: u8) -> Result<ASTNode, ParserError> {
+        // Bound recursion so deeply nested input (e.g. `=((((...))))`) returns
+        // an error instead of overflowing the stack.
+        self.depth += 1;
+        if self.depth > MAX_PARSE_DEPTH {
+            self.depth -= 1;
+            return Err(ParserError {
+                message: format!("Formula nesting too deep (max {MAX_PARSE_DEPTH})"),
+                position: Some(self.position),
+            });
+        }
+        let result = self.parse_bp_inner(min_precedence);
+        self.depth -= 1;
+        result
+    }
+
+    fn parse_bp_inner(&mut self, min_precedence: u8) -> Result<ASTNode, ParserError> {
         let mut left = self.parse_prefix()?;
 
         loop {
--- a/crates/formualizer-parse/src/tests/parser.rs
+++ b/crates/formualizer-parse/src/tests/parser.rs
@@ -4141,4 +4141,36 @@ mod r1c1_disambiguation {
             assert_not_table(formula, &spanned);
         }
     }
+
+    #[test]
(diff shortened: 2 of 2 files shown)


PSU3D0 (changes requested):
Thank you for this, and welcome. The crash is real, the mechanism you chose is the right one, and the small-stack regression test is a good idea. I reproduced the abort on main with your exact repro and confirmed the branch returns `Err` for every recursion shape I could construct, including nested arrays, range operators and mixed paren/call chains. Parser tests (330), the full eval suite (2865), clippy and fmt are all green on your head. The blast-radius framing is accurate too: `parse` is exported to untrusted strings from both the Python and WASM bindings, so this was a remote process abort.

Requesting changes on one calibration point and a few small things.

**The constant rejects Excel-legal formulas.** `MAX_PARSE_DEPTH` counts `parse_bp` frames, not nesting levels, and the top-level call plus the right-hand side of every infix operator each consume a frame. Measured on your branch: nested `IF(A1>0, …, 0)` parses to 62 levels and fails at 63; nested parens and `SUM(SUM(…))` stop at 63; right-nested arithmetic like `=1+(2+(3+…))` stops at 32 because each level costs two frames. Excel's documented limit is 64 nested function levels, so 64 nested IFs must parse. Please raise the constant so 64 function levels and 64 paren levels both parse with headroom. On a 1 MiB debug-build stack, main survives 90 nested IFs and 120 nested parens before overflowing, so 96 is a safe choice and still keeps your 1 MiB test meaningful. If you want the right-nested infix shape to reach 64 as well, that needs roughly 132, which is past the safe budget, so a code comment documenting that infix nesting costs two frames is the honest alternative.

**Pin the accepted boundary, not just the rejected one.** The test asserts 40 parens parse and 5000 fail. Please add assertions that 64 nested `IF(A1>0, …, 0)` and 64 nested parens parse, so the calibration cannot regress silently.

**Test placement.** The new test landed inside `mod r1c1_disambiguation` (tests/parser.rs around line 3972) because it was appended at end of file. Please move it to a general parser-hardening spot.

**CHANGELOG.** Add a line under `## Unreleased` → `### Security and hardening`.

Not needed from you: I filed #411 for a residual overflow this guard cannot catch (recursive `Drop` of a flat 10k-term chain), which is pre-existing and separate. I will approve the CI runs as soon as the update lands; this is on the 0.9.0 list.

PSU3D0:
@chiliec We are preparing the 0.9.0 release candidate and keeping this parser hardening on the release path. We will prepare the reviewed calibration, accepted-boundary tests, test relocation and changelog follow-up on a separate branch preserving your original commit and authorship. Please let us know by 2026-09-06 14:00 UTC if you are actively updating this so we can incorporate your work rather than duplicate it. We are also rechecking the residual #411 boundary separately; this PR will not be described as eliminating all stack-exhaustion paths.

PSU3D0:
@chiliec The tested, attributed completion is now draft #439 (ac0b8551). Your original author commit is preserved. Independent testing rejected 96 frames as unsafe on a 1 MiB stack; the final 72-frame guard passes accepted-64/rejected-5,000 coverage in debug and release on Rust 1.93 and 1.92, plus actual native Python, WASM, C ABI and Pyodide artifacts. The default Pyodide CI smoke now includes the depth matrix.

The separate flat-AST destruction risk #411 remains explicit; this is not a complete hostile-input-safety claim. Fresh PR CI is running. The September 6, 14:00 UTC checkpoint is unchanged, and your branch has not been overwritten; feedback or active follow-up is welcome before integration.

PSU3D0:

PSU3D0:
Integrated through #439 with the original author contribution preserved, the independently calibrated 72-frame limit, and native/WASM/C ABI/Pyodide regression coverage. The residual flat-AST destruction limitation #411 remains explicit. Closing this superseded PR rather than merging the older head; thank you for the contribution.

Issue #434, opened 2026-09-05
fix(eval): decouple internal mutation capture from audit-log retention


PSU3D0:
## Problem

Logged Engine mutation and atomic-action correctness currently depend on the caller's audit-log retention policy. `edit_with_logger` and `action_atomic_impl` recover newly emitted events from `log.events()[start_len..]`, but `ChangeLog::record` can discard events when disabled or evict events under a capacity limit. `ChangeLog::default()` is disabled; `ChangeLog::new()` is enabled.

With disabled capture, the internal event batch is empty. With a saturated cap, newly retained events can fall before `start_len`, so length-based slicing can miss the operation. Those events are needed for Arrow mirroring, rollback/journal construction, and topology/data invalidation—not only optional audit history. Merely bumping a clock after detecting loss cannot reconstruct a correct rollback.

This source-backed capture defect was identified during independent review of #432. That issue's focused fix handles invalidation with complete capture; it does not repair lossy capture or general inverse-replay semantics (#301). The disabled/capped numerical regression matrix still needs to be added and executed; this report does not claim runtime results for those configurations.

## Bounded correction

Capture each operation completely in private storage (or use scoped capture that defers eviction), then publish audit events honoring the caller's enabled flag, metadata, compound structure, and retention cap. Do not reject all capped or disabled logs: these are supported logging policies, not editing restrictions. Avoid a general undo-architecture rewrite.

## Required regression coverage

- Direct logged formula reordering and value-only edits with disabled/default, zero-cap, saturated-cap, and normal enabled logs.
- Multi-event atomic success, rollback, and journal undo/redo with identical graph/Arrow values and invalidation behavior independent of external retention.
- Compound boundaries and metadata; public retained logs still obey enabled/cap settings.
- Complete-capture value-only edits retain legitimate static schedule hits; formula/topology edits invalidate retained plans and schedules.

Workbook interactive defaults use enabled uncapped logs, while disabled Workbook logging normally chooses nonlogged Engine APIs. That limits the default Workbook exposure but does not make the public Engine contract sound. Until fixed, document that logged Engine correctness requires complete event capture.

PR #435, opened 2026-09-05
fix(eval): invalidate retained schedules for logged topology edits

### LINKED ISSUE #432: eval: logged formula edits and journal replay reuse stale schedule order
## Reproduced wrong values on current main

During the Astra investigation for #427, a correctness oracle reproduced stale schedule reuse through supported logged-edit and journal APIs on `8a912b50` (after #431). This is separate from the proposed performance changes; no production optimization was applied to produce it.

Use FormulaPlane Off and serial evaluation. Set `A1=1`, `B1=A1+1`, `C1=B1+1`, then evaluate all to warm the static schedule for B1/C1. In one `Engine::action_with_logger` operation, change B1 to `C1+1` and C1 to `A1+10`. Both candidates are dirty and have the same IDs as before, but their dependency order has reversed.

The next `evaluate_all` reports a static schedule cache hit and returns **B1=4, C1=11**, rather than **B1=12, C1=11**. Calling public `mark_topology_edited()` immediately before evaluation produces the correct answer, as does a freshly constructed engine containing the final formulas.

| Supported operation | Cached result (B1, C1) | Uncached and fresh-engine oracle |
| --- | --- | --- |
| `action_with_logger` formula edits | (4, 11) | (12, 11) |
| `action_atomic_journal` commit | (4, 11) | (12, 11) |
| Journal undo, after warming the correct reversed schedule | (2, 13) | (2, 3) |
| Journal redo, after warming the correct original schedule | (4, 11) | (12, 11) |

The reproducer independently resets the schedule before undo and redo so those failures are not merely consequences of the first bad commit. Built and reproduced with Rust 1.93.0; no timing claims are involved.

## Suspected boundary

`create_evaluation_schedule` keys its retained entry on the Engine topology epoch and the exact candidate vector. `edit_with_logger` mirrors recorded changes without an evident data/topology invalidation, and non-name journal undo/redo currently mark data edited rather than topology edited. Low-level graph edge changes do not automatically advance the Engine epoch.

Audit all forward, rollback and replay event boundaries. Preserve legitimate value-only schedule reuse; do not hide the defect by keying all schedules to the data snapshot. Regression coverage needs changed ordering with identical candidates, formula/value transitions, rollback, undo/redo, and the associated plan/cache invalidation contract.

## Release disposition

Treat this as a focused release blocker because ordinary supported logged formula edits can silently return stale values. Keep it separate from both #427's optimization and the broader undo tranche (#301, #303, #367, #412). Do not expand cache eligibility until the invalidation contract is verified.

PSU3D0:
## Summary

Closes #432. Logged formula edits, atomic actions, and undo/redo could reuse a static schedule after its dependency ordering changed, returning stale intermediate values. An exhaustive event classifier now publishes topology or data invalidation at the existing mutation boundaries, preserving legitimate value-only schedule hits and clearing snapshot-bound lookup indexes.

Replay and rollback invalidate before applying mutations, so a partial replay error cannot leave a changed graph behind a valid retained plan. Classification is direction-aware: undoing a value write with no old value/formula removes a vertex, and replaying it can recreate one. FormulaPlane demotion and structural boundaries retain one-batch epoch deduplication.

No schedule eligibility, cache key, ownership, default FormulaPlane mode, or general undo semantics are changed.

## Validation

- 23 focused regressions pass, including independently warmed forward/undo/redo numerical oracles, value-only cache-hit controls, retained-plan rejection, empty-precedent removal/recreation, partial inverse failure, Arrow-only structural changes, FormulaPlane epoch deduplication, and lookup snapshot invalidation.
- Full eval library suite: 2,930 passed (existing ignored tests remain ignored).
- Independent final review: clean scoped disposition; 23 focused tests and two schedule-cache controls re-run successfully.
- Parent re-ran all 23 focused tests on the committed source; formatting, Clippy and diff checks pass.

## Explicit limits

Correctness here requires complete event capture. Disabled/capped external `ChangeLog` retention can currently lose the very events needed for internal mirroring, rollback, and invalidation; that separate capture defect is tracked in #434. This PR does not repair general inverse semantics (#301) or make partially failed replay transactionally recoverable.

### DIFF
--- a/bindings/python/tests/test_inspection.py
+++ b/bindings/python/tests/test_inspection.py
@@ -254,7 +254,9 @@ def test_to_dict_full_nested_structure_for_every_report_kind() -> None:
     sheet.set_formula(1, 2, "=A1+SUM(A1:A2)")
     sheet.set_formula(1, 3, "=B1+1")
     wb.evaluate_all()
-    stamp = {"mutation_revision": 6, "recalc_epoch": 1}
+    # Logged formula topology edits now advance the mutation stamp as well.
+    # This exact fixture value checks serialization, not a one-tick-per-edit API.
+    stamp = {"mutation_revision": 8, "recalc_epoch": 1}
 
     assert wb.inspect_cell("S!B1").to_dict() == {
         "stamp": stamp,
--- a/crates/formualizer-eval/src/engine/eval.rs
+++ b/crates/formualizer-eval/src/engine/eval.rs
@@ -1708,7 +1708,6 @@ where
                 .clear_computed_overlay_after_row(sheet, before0 as usize);
             self.engine
                 .record_formula_plane_structural_change(StructuralScope::Region(affected_region));
-            self.engine.mark_topology_edited();
             if let Some(undo_ptr) = self.arrow_undo {
                 unsafe { &mut *undo_ptr }.record_insert_rows(sheet_id, before0, count);
             }
@@ -1784,7 +1783,6 @@ where
                 .clear_computed_overlay_after_col(sheet, before0 as usize);
             self.engine
                 .record_formula_plane_structural_change(StructuralScope::Region(affected_region));
-            self.engine.mark_topology_edited();
             if let Some(undo_ptr) = self.arrow_undo {
                 unsafe { &mut *undo_ptr }.record_insert_cols(sheet_id, before0, count);
             }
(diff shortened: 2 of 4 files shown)


PSU3D0:
CI found one stale Python serialization fixture: the corrected logged topology invalidation advances the inspection mutation stamp from 6 to 8 in that exact workbook. The stamp is a state revision, not a promised one-tick-per-edit counter. Commit a01945f8 updates the exact expected fixture without changing production code or weakening the nested report assertions.

All 14 inspection tests pass locally against the actual native wheel built by this PR’s CI (run 33974152890, wheels-ubuntu-latest), installed into an isolated Python 3.12 environment. Fresh CI is running on the new head; merge remains gated on green checks.

Issue #422, opened 2026-09-04
perf(explore): bound range chunk discovery and project only needed lanes


PSU3D0:
## Status and scope

Exploratory performance direction from a read-only generalist architecture audit at `79a567db55a6646c55dd07e1f99498aaee67d61a`. Source reachability was inspected; no current-HEAD benchmark or speedup was established. First measurement-and-fix tranche, after the first three candidates; begin with bounded chunk lookup.

Tracking issue: #418

## Observed implementation

`RowChunkIterator` begins at chunk index zero and scans the chunk-start list, skipping non-overlapping chunks without terminating beyond the view. Included columns construct numeric, boolean, text, error, and tag slices regardless of the caller. Numeric traversal then resolves chunk information again. Missing physical lanes have lazy null caches, so not all null-array costs are repeated; absent columns/chunks still have separate allocation paths.

## Investigation / smallest change first

- Binary-search the first intersecting chunk and stop after the last; carry the resolved chunk index into consumers.
- Measure a lane-projected typed cursor rather than always constructing generic all-lane ChunkSlices.
- Handle logical empty regions without allocating unused typed arrays; separate this from semantic changes.

## Measurement and acceptance

- Hold a tiny tail range fixed while increasing total sheet chunks; count metadata chunks visited and arrays/allocations produced.
- Compare numeric-only, mixed, wide sparse, beyond-physical-bounds, and cold/warm null-provider cases.
- Use direct SUM, lookup scan fallback, and representative aggregate readers as end-to-end controls.
- Record commit, release configuration, runtime, workbook mode, FormulaPlane mode, fixture size/hash, and repeated-sample distributions. Separate cold setup, warm evaluation, and edit costs; include correctness and relevant work/allocation counters. If the hypothesis is not material, record that outcome rather than forcing a rewrite.

## Correctness and resource guardrails

- Preserve inclusive bounds, relative row offsets, missing physical rows/columns, user/computed overlay precedence, and errors.
- Retain cancellation behavior and coordinate with #254; do not silently discard traversal failures.

## Source evidence

- [`crates/formualizer-eval/src/engine/range_view.rs:86`](https://github.com/PSU3D0/formualizer/blob/79a567db55a6646c55dd07e1f99498aaee67d61a/crates/formualizer-eval/src/engine/range_view.rs#L86)
[...]

Issue #414, opened 2026-09-02
eval: approximate VLOOKUP/HLOOKUP clone the whole lookup column on every call (~20 ms per lookup at 200k rows)


PSU3D0:
## Summary

The approximate (`range_lookup=TRUE`) branch of VLOOKUP/HLOOKUP materializes the entire lookup column into a fresh `Vec<LiteralValue>` with a clone per row on every call (crates/formualizer-eval/src/builtins/lookup/core.rs, the approximate arm around the `binary_search_match` call), then `SearchedVector::new` scans it again. Per-call cost is O(n) before the binary search even starts.

## Measured (release, evaluate_all, main at 9ff755a5)

| table rows | formulas | wall time |
|---|---|---|
| 200,000 | 20,000 | 402 s |
| 200,000 | 2,000 | 39 s |
| 10,000 | 20,000 | 15 s |

That is about 20 ms per approximate lookup at 200k rows. Exact-mode lookups on the same range go through `LookupIndex` and do not pay this.

## Fix direction

Extend the warmed lookup cache (crates/formualizer-eval/src/engine/lookup_index_cache.rs) with a sorted-projection variant keyed the same way as the exact index, so repeated approximate lookups over the same range share one projection and one sortedness verdict. That also makes the sortedness guard from #399 / #283 free.

Found while timing #399.

drafted with love by (agent) Manfred, with approval from Frankie

PSU3D0:
## Generalist performance audit follow-up — #418

Reusing this issue rather than opening a duplicate for approximate lookup setup. A read-only audit at `79a567db55a6646c55dd07e1f99498aaee67d61a` independently found the full-column materialization and `SearchedVector` projection scan. This audit ran no benchmarks; the measurements in the issue body belong to its stated earlier SHA and are not being presented as current-HEAD results.

### Additional exploratory directions

- Keep cold projection construction separate from warm search cost. O(log N) binary search does not imply O(log N) end-to-end calls if projection is rebuilt every time.
- Inspect explicit XLOOKUP `search_mode=2/-2` exact-mode paths: current control flow falls into a linear scan rather than exploiting a searchable ordered axis.
- Inspect reverse exact XLOOKUP: the lookup cache already exposes `find_last_exact`, while this mode takes a scan path. Routing eligible reverse queries through it is a bounded follow-up candidate, subject to duplicate/type/error parity.
[...]

PR #436, opened 2026-09-05
perf(eval): bound range traversal and project numeric/error lanes

PSU3D0:
## Summary

Advances #422 with a private bounded chunk cursor and numeric/error-only projection. Public `ChunkSlice` fields, row/column order, overlap, cancellation, and overlay behavior are preserved. No lookup semantics, FormulaPlane defaults, scheduler, used-extent policy, or global cache changes.

This is a measured subset, not a whole-column or general workbook speedup. Other typed lanes retain the generic adapter; sparse gaps are still traversed, and #414 remains open.

## Evidence

The inspected [report](https://github.com/PSU3D0/formualizer/blob/f44d89023436282bb8c4cfc5ec8aacecc9926bdc/benchmarks/perf-ranges-422.md), counter-free example, baseline/A source patches and summarizer are included. Raw captures, binaries and archives are not committed.

Balanced release comparisons used an exclusive build-free window, separate Cargo targets, Rust 1.93.0 and a shared Ryzen 9 3900XT host. Counter-free warm medians include:

| Fixture | Baseline | Fixed |
| --- | ---: | ---: |
| Eight-row numeric tail, default 32 chunks | 1.613 us | 1.188 us |
| Dense 32768 × 8 SUM-like traversal, default chunking | 68.030 us | 51.445 us |
| Same dense data, fragmented chunking | 270.096 us | 89.705 us |
| Million-row sparse full-span SUM-like traversal | 12.054 ms | 1.469 ms |

Direct SUM-like traversal is not general builtin SUM. Actual Engine SUM/COUNT first-evaluation results are reported separately. Each distribution has 14 fixture batches from two process runs; dispersion, cold-state limits and shared-host noise are disclosed.

There is a small accepted generic head-only regression: approximately 34 ns / 9% in the fragmented control, with smaller/default-layout effects noisier. No claim that every consumer improves.

## Validation

- Seven independent segmentation/projection, overlay, cancellation, sparse/structural/lifetime and bit/error-order regressions pass.
- Debug eval/workbook: 3,065 passed, 17 ignored.
- Release eval: 2,914 passed, 16 ignored, excluding the separately baseline-reproduced `test_scalar_arena_float_overflow` missing-panic failure.
- Targeted Clippy, formatting and staged diff checks pass.
- Independent Astra implementation/evidence review: no scoped blocker; parent verified frozen source hashes and reran the public summarizer against retained captures.

Future performance work does not automatically enlarge the frozen 0.9.0 release candidate. CI on this PR remains the integration gate.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -6,3 +6,5 @@ docs-site/src/generated/functions-meta.json linguist-generated=true
 
 # Unified diffs encode empty context lines as one space.
 benchmarks/*-baseline.patch whitespace=-blank-at-eol
+# This experimental diff also ends on an encoded empty context line.
+benchmarks/perf-ranges-422-a-only.patch whitespace=-blank-at-eol,-blank-at-eof
--- a/benchmarks/perf-ranges-422-a-only.patch
+++ b/benchmarks/perf-ranges-422-a-only.patch
@@ -1,0 +1,272 @@
+--- a/crates/formualizer-eval/src/engine/range_view.rs
++++ b/crates/formualizer-eval/src/engine/range_view.rs
+@@ -102,13 +102,20 @@
+     pub cols: Vec<ChunkCol>,
+ }
+ 
+-pub struct RowChunkIterator<'a> {
++struct RowSegment {
++    chunk_idx: usize,
++    chunk_offset: usize,
++    row_start: usize,
++    row_len: usize,
++}
++
++struct RowSegmentIterator<'a> {
+     view: &'a RangeView<'a>,
+-    current_chunk_idx: usize,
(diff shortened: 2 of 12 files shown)


PSU3D0:
All executable CI checks except the public API snapshot check passed. The snapshot difference is limited to three lines of equivalent lifetime notation: `impl<'a> Iterator for RowChunkIterator<'a>` is now rendered as `impl Iterator for RowChunkIterator<'_>`. The iterator type, item and method contract are unchanged.

I am regenerating and inspecting only that snapshot representation, keeping the measured production/harness source frozen. This is not an API-removal waiver; fresh CI remains required before merge.

Issue #411, opened 2026-09-02
parse: recursive Drop of a deep left-spine AST overflows the stack on flat ~10k-term formulas (not covered by the depth limit)


PSU3D0:
## Summary

`formualizer_parse::parser::parse` still aborts the host process on a flat, left-associative chain of roughly 10,000 terms (`=A1,A2,…`, `=1+1+…`, `="a"&"a"&…`) on a 1 MiB thread stack, and around 60,000 terms on 8 MiB. The parser itself returns `Ok` for these shapes; the overflow happens in the recursive `Drop` of the deep left-spine `ASTNode`. Replacing `drop(ast)` with `std::mem::forget(ast)` makes n=20000 succeed for every separator, which isolates the cause.

## Why it matters

This is the same untrusted-input surface as the recursion-depth DoS fixed by #408: `parse` is exported to arbitrary strings through the Python binding (`bindings/python/src/lib.rs`) and the WASM binding (`bindings/wasm/src/lib.rs`). A depth counter on `parse_bp` cannot catch it, because a flat chain consumes no parser depth by design. The threshold (~58 KB of formula text) is above Excel's 8192-character formula limit, so no workbook reaches it, but the bindings do.

## Fix direction

Either an iterative `Drop` for `ASTNode` (unroll the left spine into a worklist), or a total-length / total-node cap on `parse` input that is documented as the supported bound. The first is the proper fix; the second is a cheap interim.

Found during the review of #408.

drafted with love by (agent) Manfred, with approval from Frankie

PR #439, opened 2026-09-05
fix(parse): complete calibrated nesting guard and binding coverage

PSU3D0:
## Coordination and attribution

Draft completion of #408, retaining chiliec/Vladimir Babin's original author commit as `188ce21f`, followed by reviewed calibration and durable binding regressions. The contributor checkpoint remains **2026-09-06 14:00 UTC**; the original branch has not been overwritten.

## Scope

The parser now bounds recursive Pratt frames at 72 and returns a normal parser error for excessive nesting. This is not an exact Excel function-nesting count. Independent investigation found that the initially proposed 96-frame boundary could still abort on a 1 MiB stack, so the final boundary was calibrated and tested rather than assumed safe.

Accepted 64-level parentheses/SUM/IF and rejected 5,000-level parentheses, unary, nested functions, right-infix, arrays and power shapes are covered. Test-only code is correctly gated and relocated.

**#411 remains open:** long flat left-associative ASTs can exhaust the stack during recursive destruction. This PR does not claim complete hostile-input safety, protection on arbitrarily small host stacks, or repairs to all AST consumers.

## Validation

- Final 72-frame source: 331 parser tests pass in both debug and release on Rust 1.93.0 and stable 1.92.0, including the 1 MiB stack boundary checks. Clippy/format checks pass.
- Actual built native Python wheel: 13 focused tests pass in an isolated CPython 3.13 environment.
- Built WASM Node module: two tests pass; C ABI Rust integration: one test passes.
- External ctypes checks against the built cdylib and the existing C header/compiler smoke both pass, including returned-buffer ownership.
- Genuine Emscripten wheel in Pyodide 0.29.3: the depth matrix and existing workbook/XLSX smoke pass. The matrix is now part of the repository's default Pyodide smoke, not a temporary extra-argument invocation.
- Independent final security review: no blocker for this narrow scope. Raw wheels, logs and binaries remain local; only source tests/scripts are committed.

## Release prerequisite

This changes the payload already published as parser 3.1.0. Common and parse must move together to 3.1.1 and complete their separately authorized publication before final product 0.9.0 preflight can pass. No version bump, release tag or registry publication occurs in this PR. Focused runtime evidence is not a substitute for the final frozen-candidate package preflight.


PSU3D0:
The owner has now authorized integration ahead of the previously announced September 6 14:00 UTC checkpoint. We have refreshed the original contributor head and feedback. I am resolving the current integration conflict against cleaned main, retaining the original author commit, the reviewed 72-frame limit and the explicit #411 limitation, then rerunning current checks before merge.

Issue #427, opened 2026-09-04
perf(explore): reuse legacy topology and target closure without full rescans


PSU3D0:
## Status and scope

Exploratory performance direction from a read-only generalist architecture audit at `79a567db55a6646c55dd07e1f99498aaee67d61a`. Source reachability was inspected; no current-HEAD benchmark or speedup was established. Architectural locality/reuse direction; separate current legacy paths from already-cached mixed topology.

Tracking issue: #418

## Observed implementation

The single legacy schedule cache requires an identical candidate vector and topology epoch, deep-clones the schedule on a hit, and excludes compressed-range/dynamic candidates. Compatibility-plan reuse instead scans all schedule units and filters dirty vertices. Targeted legacy evaluation reconstructs precedent closure, including clean ancestors, on each request; target-plan execution returns through that recipe rather than retaining a legacy demand schedule. Current mixed topology has separate caching, and repeated-chain cache hits already exist.

## Investigation / smallest change first

- Measure shared immutable schedules instead of deep clones and bounded multi-entry reuse for alternating dirty branches.
- Explore topology/rank reuse independent of exact dirty masks, plus sparse projection into execution units.
- Explore reusable target closures keyed by topology/semantic revisions; only add clean-root pruning after proving invalidation closure for all relevant mutation paths.

## Measurement and acceptance

- Compare identical repeated edits, alternating branches, tiny versus broad dirty sets, deep chains, and compressed-range cases.
- Repeat a clean terminal-target request, then edit near its terminal while increasing clean ancestry depth; measure traversal visits separately from computed cells.
- Measure cache eligibility/hits, SCC/layer construction, clone allocations, full-plan scanning, retained bytes, and target recipe overhead.
- Record commit, release configuration, runtime, workbook mode, FormulaPlane mode, fixture size/hash, and repeated-sample distributions. Separate cold setup, warm evaluation, and edit costs; include correctness and relevant work/allocation counters. If the hypothesis is not material, record that outcome rather than forcing a rewrite.

## Correctness and resource guardrails

- Preserve names/tables, dynamic dependencies, cycles/SCC ordering, volatile handling, and topology/provider/symbol invalidation.
- Dirty state is not universally transitively propagated through every internal mutation path; indiscriminate clean-node pruning can be wrong.
- Bound and account new retained indexes; no per-cycle mixed-schedule rebuild claim is being made here.

## Source evidence

[...]

PR #441, opened 2026-09-05
perf(eval): share immutable cached legacy schedules

PSU3D0:
## Scope

Advances #427 with immutable sharing of the existing single-entry legacy schedule cache. Ineligible schedules remain inline; cache keys, eligibility, algorithms, target routing and FormulaPlane defaults are unchanged. One-time vector compaction avoids retaining builder spare capacity.

This is future performance work, deliberately outside the provisional 0.9.0 candidate scope.

## Evidence

[Inspected report and reproduction](https://github.com/PSU3D0/formualizer/blob/fa5c22c0/benchmarks/perf-recalc-427.md). The measured source is preserved at `4eee5aa3`, against public `8f7c7338`; both contain the same #432 correctness correction. Later integration adds current main without relabeling those measurements as a new combined-main run. Parent reconstructed current-main-plus-A and verified the merged Engine source matches exactly.

- Separate untimed controls: four depth-256 edits eliminate four nested schedule clones, modeled as 1,032 copied buffers/36,864 bytes. Measured retained-payload estimates differ by +24 bytes per entry; this is not allocator/RSS evidence.
- Counter-free B/F/F/B: 588 process samples across 21 cases, 516,180 requests. Same-chain depth-32/depth-256 observed medians change 23.900→22.926 us and 161.275→150.074 us.
- Unchanged controls also shift, so the whole 4–7% difference cannot be uniquely attributed to removed copies. Tiny alternating work is 1.53% slower. The report retains distributions, sample-unit limits and the disclosed final 11-sample timeout continuation.
- Ordinary target preparation/closure work remains unresolved. The roughly 19 ms ordinary-target versus 13 us retained-plan observation is a future investigation lead, not a measured phase breakdown or a target-cache improvement.

## Validation and limits

Independent review found no A-only blocker and reconciled all source/binary identities and 588 sample controls. The common-corrected A source passed 2,936 debug eval tests (15 ignored), 23 shared invalidation regressions, seven schedule-cache tests and 19 retained-plan controls. Parent re-ran corrected-source Clippy with warnings denied and the public summarizer. Fresh CI validates the integrated head.

Only source and inspected text are added: fixture/examples, scoped instrumentation, report, minimal baseline patch and standard-library summarizer. No raw captures, archives, binaries, private manifests or environment dumps are committed. The patch-only whitespace exception preserves an encoded trailing context line.

The alternating timed checksum covers its first branch and total computed count; separate unit tests cover both edited branches. Future zero-test subprocess hardening and a two-branch timed checksum remain explicitly nonblocking follow-ups. No new timing was run during publication preparation, and no release tag or package is published.


PSU3D0:
The Rust CI job stopped at formatting: two assertions in the untimed invalidation example needed wrapping. That test-source-only correction is now pushed; production scheduling and the counter-free timing fixture are unchanged.

Formatting now passes, and the release-built oracle was rerun on the integrated source: logged/atomic/redo yield (12,11), undo yields (2,3), matching both fresh and independently invalidated controls with schedule misses. Fresh CI is required before merge. The report continues to reproduce the original measured pair at immutable commit 4eee5aa3.

PR #461, opened 2026-09-10
fix(stats): compute normal CDF and small-argument ERF to double precision

### LINKED ISSUE #458: [bug] NORM.S.DIST and other normal-CDF functions are accurate to only ~7 decimal places
## Summary

`std_norm_cdf` uses the Abramowitz & Stegun 7.1.26 approximation, whose documented maximum error is 1.5e-7 on erf — up to 7.5e-8 on Φ. Excel's `NORM.S.DIST` is accurate to double precision, so any result that depends on the 8th decimal disagrees with Excel. Every function that calls the helper inherits the error: **`NORM.S.DIST`, `NORM.DIST`, `LOGNORM.DIST`, `GAUSS`, `Z.TEST`**.

https://github.com/PSU3D0/formualizer/blob/dcf82c64d62e73ae8d9a0e4bb37f71d28b5ea04c/crates/formualizer-eval/src/builtins/stats/mod.rs#L2853-L2871

## Environment

- Formualizer version: 0.9.2 (also 0.9.0)
- Host: WASM (npm `formualizer`)
- OS: macOS 26.6.2 (arm64)
- Compiler / Python / Node version: Node v26.4.0; also reproduced in-browser
- Enabled features (if Rust): n/a

## Reproduction

1. Create a workbook and evaluate `=NORM.S.DIST(1,TRUE())`.
2. Compare with Φ(1) = 0.8413447460685429.

```js
import { Workbook } from "formualizer";

const wb = new Workbook();
wb.addSheet("S");
const s = wb.sheet("S");
s.setFormula(1, 1, "=NORM.S.DIST(1,TRUE())");
wb.evaluateAll();
console.log(s.readRange(1, 1, 1, 1)[0][0]); // 0.8413447361676363 — expected 0.8413447460685429
```

### Minimal input

| Formula | Formualizer 0.9.2 | Exact Φ(x) | Error |
|---|---|---|---|
| `=NORM.S.DIST(0,TRUE)` | 0.5000000005 | 0.5 | 5.0e-10 |
| `=NORM.S.DIST(1,TRUE)` | 0.8413447361676363 | 0.8413447460685429 | 9.9e-9 |
| `=NORM.S.DIST(2,TRUE)` | 0.9772499371127437 | 0.9772498680518208 | 6.9e-8 |
| `=NORM.S.DIST(-2,TRUE)` | 0.022750062887256395 | 0.02275013194817921 | 6.9e-8 |
| `=NORM.S.DIST(3,TRUE)` | 0.9986500327186852 | 0.9986501019683699 | 6.9e-8 |

Reference values are Φ(x) from a 50-digit Taylor series of erf, and agree with Python's `math.erfc`. Over 15 points from x = −3 to 3, Formualizer's output matches A&S 7.1.26 **bit for bit** — including `NORM.S.DIST(0,TRUE) = 0.5000000005`, which comes from the five coefficients summing to 0.999999999 rather than 1, so erf(0) = 1e-9.

## Expected behavior

Φ accurate to double precision (about 1e-16 absolute), as in Excel. `NORM.S.DIST(0,TRUE)` should be exactly 0.5 — as the function's own docs page states.

## Actual behavior

Absolute error up to about 7e-8 — roughly seven correct decimal places.

## Additional context

**Impact.** We hit this in an option valuation whose output is `ROUND(…, 8)`. The CDF error moved the unrounded value to the other side of a rounding boundary: Formualizer returned 0.08458120 where Excel, evaluating the identical formula, returned 0.08458121 (the exact value is 0.0845812065…). Because the call value scales with the underlying price, for options worth a few dollars the 7th decimal differs as well, not just an occasional 8th.

**Suggested fix.** Replace the approximation with a double-precision one. Hart (1968), in the form given by West (2005), *Better approximations to cumulative normal functions*, has about 1e-16 absolute error across the whole real line — we checked it against a 50-digit reference from x = 0 out to ±37, including both sides of its branch switch at |x| ≈ 7.07. We are currently working around this by registering a replacement with `registerFunction("NORM.S.DIST", …, { allowOverrideBuiltin: true })`, which fixes the case above.

<details>
<summary>The implementation we register (TypeScript)</summary>

```ts
/** Standard normal CDF, Φ(x) — Hart (1968) via West (2005). */
export function normalCdf(x: number): number {
  if (Number.isNaN(x)) return Number.NaN;
  const a = Math.abs(x);
  let tail: number;
  if (a > 37) {
    tail = 0;
  } else {
    const e = Math.exp((-a * a) / 2);
    if (a < 7.07106781186547) {
      let n = 3.52624965998911e-2 * a + 0.700383064443688;
      n = n * a + 6.37396220353165;
      n = n * a + 33.912866078383;
      n = n * a + 112.079291497871;
      n = n * a + 221.213596169931;
      n = n * a + 220.206867912376;
      let d = 8.83883476483184e-2 * a + 1.75566716318264;
      d = d * a + 16.064177579207;
      d = d * a + 86.7807322029461;
      d = d * a + 296.564248779674;
      d = d * a + 637.333633378831;
      d = d * a + 793.826512519948;
      d = d * a + 440.413735824752;
      tail = (e * n) / d;
    } else {
      let b = a + 0.65;
      b = a + 4 / b;
      b = a + 3 / b;
      b = a + 2 / b;
      b = a + 1 / b;
      tail = e / b / 2.506628274631;
    }
  }
  return x > 0 ? 1 - tail : tail;
}
```

</details>

Note that `Φ(z) = 0.5 · erfc(−z/√2)` using the crate's own `erfc` would **not** fix this yet: its small-argument branch has the mistyped constant described below, and even once that is corrected, the `erfc_large` branch limits it to about 1e-9 in the tails. A dedicated double-precision Φ is still the better fix.

### Related: `ERF` / `ERFC` are inaccurate for |x| < 0.5 — one mistyped constant

`erf_approx` is documented as achieving "about 1e-15", and it does for 0.5 ≤ x ≤ 2.5 — but below 0.5 the error grows to 1e-7:

| x | `ERF` / `ERFC` error |
|---|---|
| 0.05 | 1.3e-8 |
| 0.137 | 3.5e-8 |
| 0.3 | 7.2e-8 |
| 0.49 | 1.0e-7 |

The cause is `Q[0]`, the constant term of the denominator:

https://github.com/PSU3D0/formualizer/blob/dcf82c64d62e73ae8d9a0e4bb37f71d28b5ea04c/crates/formualizer-eval/src/builtins/engineering.rs#L1997

It reads `2.84423748127893300e+03`, but in Cody's CALERF (netlib `specfun`, `DATA B`) it is `2.84423683343917062E03`. The other eight coefficients match exactly. For |x| < 0.5 the function returns `x · P(x²) / Q(x²)`, which near zero is `x · P[0]/Q[0]`, so `P[0]/Q[0]` must equal erf′(0) = 2/√π. With the constant as written it is 1.1283789100814987 rather than 1.1283791670955126 — a slope error of 2.6e-7 at the origin, which produces the table above. Evaluating the source's own coefficients in double precision reproduces Formualizer's output exactly, and with only `Q[0]` corrected the branch's error falls to at most 2.2e-16 over the same range.

`erfc_direct` returns `1.0 - erf_approx(x)` for x < 0.5, so this one constant affects `ERF`, `ERF.PRECISE`, `ERFC` and `ERFC.PRECISE` alike (we measured `ERF` and `ERFC`; the other two call the same functions). Separately, `ERF(4)` and `ERFC(4)` are off by 9.1e-10, in the `erfc_large` branch.

### Minor: `NORM.S.INV`

`std_norm_inv` has a maximum error of 1.6e-9 over p ∈ {0.001, 0.025, 0.1, 0.5, 0.9, 0.975, 0.999}, measured against Python's `statistics.NormalDist.inv_cdf` (Wichura's AS 241). That is consistent with Acklam's rational approximation without its refinement step; one Halley or Newton step against an accurate Φ would bring it to double precision.

cloudexible:
Fixes #458.

## Summary

- **Normal CDF.** `std_norm_cdf` used the Abramowitz & Stegun 7.1.26 approximation, accurate to only ~7e-8. It's replaced with Hart (1968) as given by West (2005), the algorithm suggested in the issue. Absolute error is now at most 2.2e-16 over z ∈ [−40, 40], and `NORM.S.DIST(0,TRUE)` returns exactly 0.5. This fixes `NORM.S.DIST`, `NORM.DIST`, `LOGNORM.DIST`, `GAUSS` and `Z.TEST`.
- **`ERF` / `ERFC` for |x| < 0.5.** `Q[0]` in `erf_approx` was `2.84423748127893300e+03`; Cody's CALERF value is `2.84423683343917062e+03`. With the typo, `P[0]/Q[0]` was 1.1283789100814987 instead of 2/√π = 1.1283791670955126. Error in this range drops from up to ~1e-7 to under 1e-15. This fixes `ERF`, `ERF.PRECISE`, `ERFC` and `ERFC.PRECISE`.

| Formula | Before | After | Reference |
|---|---|---|---|
| `=NORM.S.DIST(0,TRUE)` | 0.5000000005 | 0.5 | 0.5 |
| `=NORM.S.DIST(1,TRUE)` | 0.8413447361676363 | 0.8413447460685429 | 0.8413447460685429 |
| `=NORM.S.DIST(1.96,TRUE)` | 0.9750021738917761 | 0.9750021048517795 | 0.9750021048517795 |
| `=GAUSS(1)` | 0.3413447361676363 | 0.34134474606854304 | 0.3413447460685429 |
| `=ERF(0.3)` | 0.32862668757478203 | within 1e-15 | 0.3286267594591274 |

Reference values come from Python's `math.erfc` / `math.erf`.

## Tests

- Regression tests added: `std_norm_cdf_is_double_precision` (central values, both sides of the |z| ≈ 7.07 branch switch, far tails, z = ±38 cutoffs, NaN) and `erf_small_argument_branch_is_double_precision`.
- Six expected values in `tests/formula_tests/extended_coverage.json` encoded the old approximation's output (for example `GAUSS(1)` expected the exact wrong value reported in the issue). I checked each new value against an independent reference before updating it.
- `cargo test --workspace --exclude formualizer-python`: 3914 passed, 0 failed.
- `cargo fmt --all -- --check` is clean. `cargo clippy --workspace --exclude formualizer-python --all-targets -- -D warnings` is clean.

## Notes for reviewers

- **Tail relative accuracy.** Hart/West is accurate to ~1e-16 in absolute terms, as the issue asks. For Φ below about 1e-6 (z < −5) its *relative* error reaches ~1e-8 (worst near the 7.07 branch switch). That's much better than before, but a follow-up could improve it if tiny p-values need full significant digits.
- **Not addressed here:** the smaller `NORM.S.INV` error (~1.6e-9) and the `ERF(4)` / `ERFC(4)` error (~9e-10) from the `erfc_large` branch, both mentioned in the issue.
- **Separate `Z.TEST` behaviour.** When sigma is omitted, `Z.TEST` uses the population standard deviation, while Excel uses the sample standard deviation. On `=Z.TEST({3,6,7,8,6,5,4,2,1,9}, 4)` that changes the result by ~0.01 (0.0793 vs Excel's 0.0906). This PR doesn't change it; the updated goldens keep the current population-SD behaviour.
- **Not run in my environment:** `formualizer-python` doesn't link without `python3.14-dev`. It also has a pre-existing clippy error in `bindings/python/src/workbook.rs` (too many arguments), which this PR doesn't touch. `xtask audit --strict` reports 368 issues both before and after this change. I didn't rebuild the WASM package.

https://claude.ai/code/session_01SnCbkedAgifUkButpq4KGM


PSU3D0:
Independent numerical review is complete: compiled helper bodies compared against 90-digit references at 70,001 grid/random points plus adjacent-float boundaries; 2,988 eval tests passed locally, and hosted binding/CI checks are green. Scoped approval for central CDF absolute accuracy and the small-argument ERF correction. Maximum sampled CDF absolute error was about 2.25e-16, so the claim is approximately machine precision, not a universal 1e-16 bound. Remaining ERFC/inverse/tail and Z.TEST convention work is preserved in #464. Thank you for the careful scope disclosure and contribution.

